PR #1, opened 2017-05-01
Add instructions to integrate Gogs with Mailcow

ecdsa521:
Simple instructions on how to integrate Gogs with Mailcow.


andryyy:
Thanks! :)

PR #5, opened 2017-05-06
fixes portainer password issues in mailcow/mailcow-dockerized#228

philip-ulrich:
password setting had no persistent storage. added persistent storage per official compose docs: https://github.com/portainer/portainer-compose/blob/master/docker-compose.yml

I believe this issue lies from the fact that originally the dockerized version of portainer used http auth, but switched to app auth. Docs were never updated.

### DIFF
--- a/docs/portainer.md
+++ b/docs/portainer.md
@@ -6,7 +6,7 @@ In order to enable Portainer, the docker-compose.yml and site.conf for nginx mus
       image: portainer/portainer
       volumes:
         - /var/run/docker.sock:/var/run/docker.sock
-        - /opt/portainer/data:/data
+        - ./data/conf/portainer:/data
       restart: always
       dns:
         - 172.22.1.254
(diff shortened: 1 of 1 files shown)


andryyy:
Hi, thanks! Can you rebase on dev? Any maybe use /data/conf/portainer as data directory in the docs? (Remember to add a mkdir etc.). :)

philip-ulrich:
There is no dev branch for this docs repo, so I'm supposing you would like me to make one? Or are you referring to making changes in mailcow-dockerized?

Is this necessary though since this is one of the add-ons and not an official part of the install? If you would like the data dir to be in the normal location, that's a simple change in the md here in the docs. Docker-compose will handle the mkdir and portainer will create whatever files it needs.

andryyy:
Oh my, I'm sorry. Will merge this when I'm home. :)

philip-ulrich:
👍 No worries. I changed the data dir for you to the normal conf dir. Probably better that way anyway.

PR #13, opened 2017-05-07
Fixed DKIM selector in example

vivernet:
In the current version the choice DKIM selector is not available. And when you create a DKIM key, it is created with a selector "dkim"


andryyy:
Hi, this will be available very soon. :) But I merge it to fit the default selector (which will indeed be dkim). Thanks!

PR #16, opened 2017-05-13
docker-compose down needed to enable IPv6

mkuron:
As requested in https://github.com/mailcow/mailcow-dockerized/pull/203, this PR documents that `docker-compose down` is needed after upgrading to the most recent version. If that isn't done, `docker-compose up -d presents the following error message:
```
ERROR: Network "mailcowdockerized_mailcow-network" needs to be recreated - enable_ipv6 has changed
```

### DIFF
--- a/docs/install-update.md
+++ b/docs/install-update.md
@@ -55,6 +55,12 @@ docker rm $(docker ps -a -q)
 
 ### Step 2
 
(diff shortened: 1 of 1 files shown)


broedli:
Thanks.
Since this isn't in master yet, I'll wait on merging this ...

PR #18, opened 2017-05-13
Seemingly unneeded tags in restore

stevesbrain:
{% raw %} and {% endraw %} appear unneeded/erroneous. Command executes fine in ```/bin/bash``` without this at least. Please ignore if this is needed in other circumstances.


andryyy:
Yes, thanks, they are not needed anymore.

PR #25, opened 2017-06-03
Added client_max_body_size option to nginx config

coin3d:
I had a problem with my iPhone and SOGo (using EAS) when it was refusing to send a mail with a bigger attachment. The reason was that my reverse proxy was refusing files bigger than the default of 1 MB, that I had configured like given in the current documentation. So I just copied the same configuration line as already configured in the SOGo vhost ([link](https://github.com/mailcow/mailcow-dockerized/blob/master/data/conf/nginx/site.conf#L102))


andryyy:
Thanks

PR #27, opened 2017-06-13
Make backup scripts cron executable

levinus:
The backup scripts of both mysql and maildir could not be executed by the cron daemon or similar services. I solved this by not allocating ttys when running `docker` / `docker-compose`.


andryyy:
Hi, do they still work within a console? I thought about adding it as a hint or something. But when they work with -T it is fine. But please test it! :)

> Am 13.06.2017 um 15:31 schrieb Levin <notifications@github.com>:
> 
> Don't

levinus:
I tested them but forgot to add it.
I first also thought of adding it as a Info but they both work also fine in a console.

andryyy:
Cool, then let's merge it.

PR #28, opened 2017-06-18
Prevents changes from being overwritten in updates

PhoenixPeca:
Ref: [#370](https://github.com/mailcow/mailcow-dockerized/pull/370)


mkuron:
It shouldn't be necessary to copy and paste the entire config. This should suffice:
```
$config['useEASforOutlook'] = 'no';
```

mkuron:
That doesn't work. You are overwriting `$config` completely, so things like the server name wouldn't be available anymore. You need to actually set individual members as I suggested before.

docs/u_e-mailcow_ui-config.md
```diff
@@ -1,4 +1,4 @@
-Several configuration parameters of the mailcow UI can be changed by creating a file `data/web/inc/vars.local.inc.php` which overrides defaults settings found in `data/web/inc/vars.inc.php`.
+Several configuration parameters of the mailcow UI can be changed by creating a file `data/web/inc/vars.local.inc.php` which overrides defaults settings found in `data/web/inc/vars.inc.php`, among others include `data/web/autodiscover.php`.
```
mkuron: Should be "including" instead of "include".

PhoenixPeca: cool thanks

PhoenixPeca:
This issue has been managed #30

PR #30, opened 2017-06-23
Remove obsolete information

mkuron:
Remove obsolete information:
- A fail2ban-like mechanism is now integrated into Mailcow. Using the syslog docker backend is therefore not supported anymore.
- `_autoconfig._tcp ` doesn't actually exist, so remove it. (googling it only reveals one other hit)
- Changes from #28, but corrected and adapted to https://github.com/mailcow/mailcow-dockerized/pull/389
- Bind was replaced with Unbound

Closes #28


PhoenixPeca:
Seems like the docs have been updated. Is it ok to close this?

mkuron:
One thing was still missing, but I've resubmitted that to #31

PR #39, opened 2017-08-19
Debugging: Docker and firewalls may have created conflicting rules

mkuron:
Docker and host-based firewalls both modify iptables rules, and both expect to be the only thing modifying them. So sometimes, conflicts can arise, resulting in blocked connections. As part of debugging blocked connections, one should turn off the host firewall to see whether it is the culprit and manually create appropriate rules to permit the connections.

Original report: https://github.com/mailcow/mailcow-dockerized/issues/504

### DIFF
--- a/docs/debug-common_problems.md
+++ b/docs/debug-common_problems.md
@@ -56,6 +56,7 @@ while trying to start / install mailcow: dockerized, make sure you've followed o
 ## XYZ can't connect to ...
 
 Please check your local firewall!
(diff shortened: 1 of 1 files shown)


andryyy:
Nice, thanks!

JiiPee74:
People who use CentOS and FirewallD should take a not of this, it's also valid for other distros using FirewallD.
https://docs.docker.com/v1.6/installation/centos/#firewalld

And this is only valid for ipv4, for ipv6 docker itself won't(?) add any rules.

PR #40, opened 2017-08-31
Correct automatic doveadm commands

strobeltobias:
The commands didn't work on my setup before I changed them to this.


andryyy:
Thanks!

PR #45, opened 2017-09-16
More client config guide fixes

mkuron:
Fixes the issue reported in https://github.com/mailcow/mailcow-dockerized/pull/468#issuecomment-329975117


andryyy:
Thanks!

PR #58, opened 2017-11-24
Remove erroneous component of "cd" instruction

stevesbrain:
Shouldn't be trying to cd into a directory that doesn't yet exist (this is created in the "mv" step just below)


PhoenixPeca:
Thanks!

PR #62, opened 2018-01-05
add instruction for outlook 2016

mritzmann:
The new Outlook 2016 Setup Assistent for EAS didn't work with Mailcow. See this Issue with explanations from @mkuron: https://github.com/mailcow/mailcow-dockerized/issues/615#issuecomment-332317644

This PR add instruction for Outlook 2016 to the docs.

### DIFF
--- a/docs/client-outlook.md
+++ b/docs/client-outlook.md
@@ -1,12 +1,12 @@
 <div class="client_outlookEAS_enabled" markdown="1">
 
-## Outlook 2016 on Windows
+## Outlook 2016 (from Office365) on Windows
 
 <div class="client_variables_unavailable" markdown="1">
   This is only applicable if your server administrator has not disabled EAS for Outlook. If it is disabled, please follow the guide for Outlook 2007 instead.
 </div>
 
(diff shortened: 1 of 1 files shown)


andryyy:
This seems to only affect Outlook from Office 365 packages. Can you mention this in your PR? :-)

mritzmann:
Agreed. I corrected it. :-)

Tschlienger:
My friend and I just set up a new mailcow server and everything works just fine. We discovered that we had to enable TLS 1.0 for Outlook to be able to use autodiscover. Even then, only the old setup (as you mentioned in your PR very nicely, thanks for that btw), the new setup agent still doesn't work on March 12th. Are there any updates about why this is happening? Or did anyone find a solution for that yet? We will have some very basic experienced users who will be clearly overwhelmed if we tell them to use the old setup.

Thanks,
Tobias

mritzmann:
@Tschlienger See also: https://github.com/mailcow/mailcow-dockerized/issues/615#issuecomment-332317644

PR #65, opened 2018-01-21
Remove everything containing rmilter because it is not used anymore

MAGICCC:
As seen in https://github.com/mailcow/mailcow-dockerized/issues/945


andryyy:
Thanks!

PR #66, opened 2018-01-21
Ensure that configuration guide params are stored URL-encoded.

icedream:
Fixes mailcow/mailcow-dockerized#913.

Safari does not support non-ASCII characters for cookies and seeing that other websites usually URL-encode their cookie values I decided to implement this into the JavaScript for the doc pages as well.


PhoenixPeca:
cool, thanks!

PR #67, opened 2018-01-28
Add instruction for pulling + starting portainer container

MAGICCC:


andryyy:
I changed some lines for a persistent configuration. You couldn't know, it wasn't documented.

PR #77, opened 2018-03-03
Fix HOW TO sogo

MAGICCC:
It seems that you need to fully recreate SOGo **and** NGinx folder to see the new image

### DIFF
--- a/docs/u_e-sogo.md
+++ b/docs/u_e-sogo.md
@@ -20,4 +20,7 @@ SOGo default:
 
 ## Change Logo
 You can change SOGo's logo by replacing `data/Dockerfiles/sogo/sogo-full.svg`.
-After you replaced said file you need to rebuild the container by executing `docker-compose build sogo-mailcow` and start it with `docker-compose up -d sogo-mailcow`.
+After you replaced said file you need to rebuild the SOGo and NGinx container by executing following command:
+```
+docker-compose up -d --no-deps --force-recreate nginx-mailcow sogo-mailcow
+```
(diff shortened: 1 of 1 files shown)


mkuron:
That is not true. The svg file is only copied into the SOGo container. We don‘t even have a Dockerfile for nginx (we use the official image), so rebuilding the image will do nothing.

You may, however, need to reset the nginx container (`docker-compose rm -s nginx-mailcow; docker-compose up -d`) as nginx caches SOGo resources for a few hours.

MAGICCC:
By rechecking it I just found out that my command is indeed wrong. I debugged this some hours in the first instance and used deprecated command. Care to unlock the PR so I can commit `docker-compose up -d --build sogo-mailcow`?

mkuron:
Of course you need to rebuild Sogo, that‘s what the documentation says already. If you want to add something on clearing the Nginx cache (my command is not the only possible one for that and definitely not the fastest way to do it), go ahead.

mkuron:
Your new commit has showed up now, but it is exactly equivalent to what the documentation previously said. You just combined the two commands into one. Please revert to what it originally said and only add something on clearing the Nginx cache.

MAGICCC:
In my recent tests I only needed to invoke `docker-compose up -d --build sogo-mailcow` in order to 'update' the logo

mkuron:
But that‘s what the commands previously in the documentation do too. So there is nothing that needs to be changed in the documentation?

MAGICCC:
Ah welp, I have **literally no** idea what I did in the last hours, but somehow it didn't update the logo, but now it does...
Well either close the PR or merge it so users only need to invoke one command

PR #81, opened 2018-04-08
Fix cronjob errors of Dovecot expunge

strobeltobias:
Recently, I discovered that the cronjob on my mailcow-dockerized host isn't working.
If the script is called via crontab, the docker-compose command doesn't know where the config file is located. So I added the "cd" to change the directory to the working directory of the mailcow-dockerized instance.
When executing the docker-compose exec command via cronjob, it fails with error message "the input device is not a TTY". By adding the parameter "-T" it works.

### DIFF
--- a/docs/u_e-dovecot-expunge.md
+++ b/docs/u_e-dovecot-expunge.md
@@ -31,8 +31,9 @@ If you want to automate such a task you can create a cron job on your host that 
 
 ```
 #!/bin/bash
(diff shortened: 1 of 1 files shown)


MAGICCC:
Note that not all users install their mailcow to `/opt/` I would rather add a note regarding this.

strobeltobias:
You're right, @MAGICCC. I thought it's fine that way as the documentation suggests it as default path. I added a note now.

andryyy:
Or use `-f /path/to/your/mailcow-dockerized/docker-compose.yml`. :-) But this approach is perfectly fine.

PR #83, opened 2018-04-19
Add release cycle information

xetorixik:
Add release cycle information as discussed > https://github.com/mailcow/mailcow-dockerized/issues/1290


andryyy:
Merci!

PR #89, opened 2018-06-10
Added Gitea third party docs

ntimo:
Hello,
I have added the docs for Gitea. Gitea is a much more advanced version of Gogs, having both setup instructions would be kind of nice :)


andryyy:
Thanks! :-)

Have you tested it on a clean install?

ntimo:
@andryyy Yes I did, it works perfectly 👍

PR #98, opened 2018-09-21
Version number changed in the link

T0biii:


MAGICCC:
Thanks 31.0.6 is some months old and nobody updated it :(

PR #100, opened 2018-09-23
Update client-thunderbird.MD (SOGo Integrator link) 

T0biii:
Edit link to SOGo Integrator because of 404 Error
Maybe fixed it?


andryyy:
Have not checked, but we cannot break the link more than it already is. Thanks. :-)

PR #101, opened 2018-09-25
added documentation to secure the installation with iptables

chilly2go:
Hi,

when first starting with mailcow I was also new to docker and wanted to secure my vps. 

It took me a many hours of searching and trial and error to find what I now want to contribute to the documentation.
The link would be sufficient but you never know how long it will be available (and people are lazy).

### DIFF
--- a/docs/firststeps-iptables.md
+++ b/docs/firststeps-iptables.md
@@ -1,0 +1,91 @@
+When running a machine which is publicly available it is mandatory to secure it.
+One step is to use a firewall like _iptables_ to block all ports that should not be opened to the public.
+
+This guide mainly copies a [blog entry at unrouted.io](https://unrouted.io/2017/08/15/docker-firewall/) and makes the essential points available here.
+There are more measurements that should be taken to secure a machine.
+
+## Make iptables rules persistent 
+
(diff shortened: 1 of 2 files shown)


mkuron:
Actually, we specifically recommend not using a host firewall when running Docker. If you look through the issue tracker, you‘ll notice that a significant fraction of the issues is due to misconfigured firewalls or Docker interacting unfavorably with iptables. 

Also, you don’t technically need a firewall if all you are running on your host is Mailcow. It has no unsecured services and no ports that are not meant to be open to the public.

chilly2go:
what I described would not interfere with docker rules (directly) but add another chain that can handle own rules.

AFAIK it is not best practise to not close unused ports. That is why I thought it important to have this info available. 

(no need to answer. just wanted to state why I did this)

mkuron:
Docker assumes it has exclusive control over iptables. Your solution may work for you right now, but things like restarting Docker could break it. If we add your stuff to the documentation, people will add their own rules too and not consider how they interact with Docker.

If there is nothing listening on a port, it doesn’t matter whether it is firewalled or not.

PR #102, opened 2018-09-26
Change wget URL to new Thunderbird60 compatible version

Xadagain:
Change the download URLs of SOGo integrator and connector to new compatible to Thunderbird 60 version. Testet


levinus:
Can confirm this works.

andryyy:
Thanks!

PR #103, opened 2018-10-03
CardDAV addressbooks in Roundcube

heavygale:
Instructions on how to integrate SOGOs CardDAV addressbooks in Roundcube - preconfigured for every user.

### DIFF
--- a/docs/third_party-roundcube.md
+++ b/docs/third_party-roundcube.md
@@ -88,6 +88,7 @@ Download the latest release of [RCMCardDAV](https://github.com/blind-coder/rcmca
 ```
 cd data/web/rc/plugins
 wget -O - https://github.com/blind-coder/rcmcarddav/releases/download/v3.0.3/carddav-3.0.3.tar.bz2 | tar xfvj -
+chown -R root: carddav/
 ```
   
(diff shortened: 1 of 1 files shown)


andryyy:
Hi, is this not stable yet?

heavygale:
Sorry, I didn't mean to close this PR, I just cleaned up my repos. Looks like that unintentionally closed the PR, too.

My tests worked well: With `$config['address_book_type'] = '';` in the roundcube config file the default address book is removed (the one that's not synced via CardDAV) and with `'use_categories' => true` in the plugin config file even the categories assigned to contacs are listet in roundcube. This way the user will see the the same contacs in roundcube as in SOGo or via CardDAV - preconfigured by the server admin. 

contacts

The only thing I couldn't figure out is, if or how it's possible to not only include the default address book created by SOGo but additional addressbooks the user created (e.g. in SOGo), too. The config file explains: "If the URL points somewhere in the CardDAV space, but _not_ to the location of a particular addressbook, the server will be queried for the available addressbooks and all of them will be added." That's what I did with `'url'  =>  'https://mx.example.org/SOGo/dav/%u/',` or even shorter variants, but it always only listed the main addressbook. But that should be a minor issue, as most users probably only use one addressbook anyway.

andryyy:
Oh, okay, can you add this as note? I will merge it then.

heavygale:
I've added this information.

heavygale:
Still works fine on my installation.

MAGICCC:
Ah well \o/

PR #105, opened 2018-10-06
Use GitHub API for docker-compose download

Knight1:
Using the GitHub API makes the project more independent and is always returning the latest version from github.com


andryyy:
No, there was another PR where we explained it.

PR #106, opened 2018-10-06
IMAP IDLE interval documentation

levinus:
Added doc about imap idle interval. Let me know if you find any mistakes. Tested with a local mkdocs installation.


andryyy:
Amazing, thanks!!

PR #110, opened 2018-10-15
LE and we prefer v6 & check tutorial

Knight1:
LE and we prefer v6 & check tutorial so that no one is confused and knows how to check that everything is setup right and do not need to wait 24 hours.

### DIFF
--- a/docs/firststeps-ssl.md
+++ b/docs/firststeps-ssl.md
@@ -30,7 +30,15 @@ Do not use quotes (`"`)!
 ADDITIONAL_SAN=cert1.example.org,cert1.example.com,cert2.example.org,cert3.example.org
 ```
 
-Each name will be validated against its IPv4 address.
+Each name will be validated against its IPv6 address (prefered if present) or IPv4.
+
+Before recreating the acme container make sure that your Domain provider actually already returns the new address(es). Some needs a bit of time to do that and when it fails the next retry is after 24 hours.
+
+```
+dig cert1.example.org NS +short | head -1
+dig cert1.example.org +short @(return from above)
+```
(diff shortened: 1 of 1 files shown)


Knight1:
The magical effect is that many providers need time to push out the changes through out the backend. 
I only know CloudFlare where it is basically instant. OVH needs up to  5 minutes, Google Cloud DNS 2-3 minutes.
If the acme container is restartet while the changes are not yet deployed to the nameservers Let’s Encrypt will fail validation. Let’s Encrypt uses the nameservers for the validation and prefer IPv6 If present. I guess that would solve some problems because people know a bit about the validation process. 

Or what do we need to change for a better understanding that the user needs to wait a few minutes?

andryyy:
Our ACME script does pretty much that. It checks wether the record is propagated and skips it if not. It does not fail validation.

Knight1:
We only do er preflight Check which is good because of Le’s limits. But we only retry it after 24hours. To prevent that waiting time and issues I want to clarify that and add the tutorial so the use can check if the records are present.

andryyy:
Ah, I see what you mean. I misunderstood you.

But why would you query the domains own NS? You should check the name against a public resolver. :-) It is still not great, but probably a bit more reliable.

I would prefer a note like "if your certificate is missing, check acme-logs for non-verified domain names and restart the container to force a retry" or something like that. People will not query their NS and say "see, it's fine, acme-mailcow is crap" then, just because their names have not propagated yet. 👍

mkuron:
> preflight check

This check is subject to the unbound container's caching. So you would beed to either remove that check or restart the container.

andryyy:
We can add cache-max-negative-ttl and cache-min-ttl (which needs to be equal or less than cache-max-negative-ttl). This would decrease the NXDOMAIN cache lifetime.

milkmaker:

PR #114, opened 2018-10-29
fix gitea instructions

pjv:
the `app.ini` needs to be created before the container is brought up and the gitea page initially accessed otherwise you get the gitea default app.ini created and it doesn't work at all.


milkmaker:

PR #121, opened 2019-01-03
Disable greylisting

pjv:
fixes https://github.com/mailcow/mailcow-dockerized/issues/413


andryyy:
Thanks, I will try to write something for the UI, too.

PR #125, opened 2019-01-05
Fix duplicate (and wrong?) mail address

jonasc:
Found this reading the documentation. If I'm not mistaken, the proposed change was how the text was intended.


andryyy:
Indeed. :) thank you!

PR #128, opened 2019-01-19
[Dovecot] Just some tiny tweaks in FTS docs

patschi:


patschi:
Silently addressed in 69e304ef753a7562fba783d7cbdf585b9f67d07c.

PR #131, opened 2019-01-26
hostname ≠ ui-domain

heavygale:
We need a additional variable for the mailcow-ui-domain, if the domain used for the mailcow ui differs from the hostname. Otherwise our users will get the login promt, when they try to access the mobileconfig file.
The mailcow-ui-domain will be passed on from the mailcow-ui, if the corresponding PR gets merged: https://github.com/mailcow/mailcow-dockerized/pull/2229

### DIFF
--- a/docs/clients.js
+++ b/docs/clients.js
@@ -3,12 +3,12 @@ if (window.location.href.indexOf('/client/') >= 0) {
         function setCookie(name, value) {
             document.cookie = encodeURIComponent(name) + "=" + encodeURIComponent(value) + "; path=/";
         }
-    
+
         function getParameterByName(name) {
             var match = RegExp('[?#&]' + name + '=([^&]*)').exec(window.location.hash);
             return match && decodeURIComponent(match[1].replace(/\+/g, ' '));
         }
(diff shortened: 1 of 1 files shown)


PhoenixPeca:
Hi,

in the "vars.inc.php" file there is a `$mailcow_hostname` variable.
If your mailcow hostname is different from your reverse-proxied mailcow ui, then you may override that variable in "vars.local.inc.php".

Note: you may also want to change other config references to that variable.

andryyy:
Would be good if it falls back to the host variable if ui is not set (not updated mailcows).

heavygale:
@PhoenixPeca 
Ok, that will do the trick for me - although I'd maybe prefer to have the hostname shown as host, except for references to the mailcow ui.

But if the reverse proxy for the mailcow ui runs on a different machine than the mailcow itself, all the host-values in the docs would get replaced with the reverse proxy's domain, too. At least in this case two seperate variables are needed, if the reverse proxy only handles the http(s) connections.

@andryyy 
Of course, I'll change that.

PR #134, opened 2019-02-27
Clarify catch-all aliases with alias-domains

evilstiefel:
The docs don't yet mention alias domains and their effect when using catch-all aliases for a domain within mailcow. This added example should clarify that any mail send to an alias domain will in fact be delivered to the catch-all alias on the target domain, which is in contrast to specific aliases added for a single mailbox.


andryyy:
I wonder if we should change that though.

PR #145, opened 2019-04-21
Add instructions for CoreOS

pazaan:
* Fixes https://github.com/mailcow/mailcow-dockerized/issues/2391


pazaan:
CoreOS has been deprecated, so not particularly relevant any more.

PR #147, opened 2019-05-04
Fixed logging driver compose example

ntimo:
The logging driver has to be driver not log_driver :)


andryyy:
Same for all docker compose versions?

ntimo:
Yes compose v3 also uses driver: as well as v2.

andryyy:
Nice, thanks for checking that. I must have missed it then.

PR #152, opened 2019-05-29
add missing ipv6 configuration for nginx SSL example

hachre:
fixes https://github.com/mailcow/mailcow-dockerized/issues/2658


andryyy:
Mörci. :)

PR #153, opened 2019-06-09
Update i_u_m_install.md

prhomhyse:
The URL of Docker Install has been changed from https://docs.docker.com/engine/installation/linux/ to https://docs.docker.com/install/

I think this is their way of putting all possible means to install Docker in one URL. 
Although the old link redirects pretty neatly, I feel it's best the URL is updated to reflect the new one to avoid a broken link later on.


MAGICCC:
Thanks!

PR #160, opened 2019-07-15
Add relayhost docs

### LINKED ISSUE #157: Clarification on relayhosts
The documentation [here](https://mailcow.github.io/mailcow-dockerized-docs/firststeps-relayhost/) tells me to edit a domain and to assign a relayhost to it. When editing a domain though, all I have is a checkbox to enable relaying of the domain, I can't "assign" a relayhost. I later found [this](https://github.com/mailcow/mailcow-dockerized/issues/1724) issue, which clarifies that checking the checkbox makes Mailcow relay mails to the primary mail service it finds by looking at the MX DNS-records of the domain.

Please clarify that
1. checking the checkbox really does all the magic and you don't have to mess e.g. with transport maps (which is what I did first)
2. Mailcow selects the destination by selecting the primary MX entry from DNS

(Note: I might create a PR for that later, but I want to get my mail setup done first)

fm:
Fixes #157

### DIFF
--- a/docs/firststeps-relayhost.md
+++ b/docs/firststeps-relayhost.md
@@ -1,3 +1,36 @@
-As of September 12th you can setup relayhosts as admin by using the mailcow UI.
+As of September 12, 2018 you can setup relayhosts as admin by using the mailcow UI.
 
-Edit a domain afterwards and assign a relayhost for it to use.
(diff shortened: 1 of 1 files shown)


andryyy:
Thank you very much!

PR #161, opened 2019-07-21
Added Telegram

patschi:


andryyy:
Hope we see @mkuron there too

PR #170, opened 2019-10-19
Create note for forece ckecking recipient domains for existence

Clevero:
In my use case, this is needed when the recipient domain is a relay domain.

More explained in https://github.com/mailcow/mailcow-dockerized/issues/2981

Since there could be more problems when this setting would be a default, so here just a note


andryyy:
This is a change, that breaks a lot of things in mailcow. It needs to address and explain that.

milkmaker:

PR #172, opened 2019-10-19
Added support to new SOGo Connector

dragoangel:
Fixes #171.

Added SOGo Connector for Thunderbird 68+.
Added 4rh parameter to display maximum supported Thunderbird version in versions.csv.
Separated sed command for each plugin to easily remove (comment-out) unneeded plugin or add new one.
My version of build-plugins.sh with additional plugin and minimum/maximumum supported TB versions parameters.

### DIFF
--- a/docs/download/build-thunderbird-plugins.sh
+++ b/docs/download/build-thunderbird-plugins.sh
@@ -10,18 +10,45 @@ fi
 
 cd $(dirname $0)
 
-wget -O integrator.tar.gz https://github.com/inverse-inc/sogo-integrator/archive/master.tar.gz
+# download SOGo Connector (TB68+)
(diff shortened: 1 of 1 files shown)


mkuron:
- Please remove all the extra extensions you are installing. We shouldn‘t unnecessarily bloat the set of extensions we install.

- Also, remove all the Thunderbird < 68 code. Thunderbird 60 is going out of support in a few weeks and there is no automated way to upgrade from Integrator/Connector 60 to Connector 68. Then we also don‘t need the maximum version stuff; there is never going to be another update to version 60.

dragoangel:
I look in all it, and thinking maybe earlier describe how-to use TbSync with CalDAV & CardDAV provider?

mkuron:
Feel free to add it to the guides; I haven’t used it myself. SOGo has the advantage that it configures itself automatically, and that‘s what we required for a client to be added to the guides.

PR #174, opened 2019-10-20
SOGo Integrator now built into Connector

mkuron:
Also, don't install any other add-ons by default.

Fixes #171. Fixes #172.

### DIFF
--- a/docs/client/client-thunderbird.md
+++ b/docs/client/client-thunderbird.md
@@ -23,34 +23,34 @@
 <li>
   Close the account settings window with the <i>OK</i> button.
 </li>
-<li class="client_integrator_enabled">
-  In your web browser, download <span class="client_variables_unavailable">SOGo Integrator</span><span class="client_variables_available"><a class="client_var_integrator_link client_var_link" href="/thunderbird-plugins/sogo-integrator-__VERSION__-__DOMAIN__.xpi">SOGo Integrator</a></span>.
+<li class="client_connector_enabled">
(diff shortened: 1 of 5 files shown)


andryyy:
Cool, thanks. Feel free to merge when you think it's complete.

mkuron:
@dragoangel, please check whether this works for you too.

andryyy:
Thanks :)

Braintelligence:
@mkuron Hey, can you confirm if it's possible to run multiple SOGo connectors? Or can I only have one domain active in that way per Thunderbird instance?

mkuron:
You can obviously only install the plugin once. It automatically configures itself based on the default IMAP account and has no way to manually configure.

If you want to use multiple users' calendars/address books, the best solution is to set up ACLs to share them with your primary user.

Braintelligence:
What if I have a mailcow running many multiple domains and I want to use multiple mail accounts separately together with their respective calendars and address books without sharing them between those accounts?

I guess I'll have to set them up the old way as remote address books and calendars, by inputting the Card/CalDAV-URL right?

mkuron:
Correct. That's just a limitation of the plugin. I tend to use sharing for this situation though, which you can even use between multiple domains.

Braintelligence:
I understand. Thank you very much for clarifying. Maybe we should put this bit into the docs for those that touched the integrator functionality for the first time, like me. I only worked with the Connector before.

PR #177, opened 2019-10-30
Made docs more concrete

jacboy:
To get access to the profile for Apple devices the user must be logged in into the Mailcow UI with the mailbox user account he wants to add to his device.


andryyy:
Thanks. If you rename Mailcow to mailcow, I will merge it. :)

jacboy:
Change has been implemented. I would like to say Thank You for making my life as an admin easier because I don't have to spend hours and hours on mailservers.

andryyy:
Thank you!! :)

PR #181, opened 2019-11-21
Update third_party-roundcube.md

Houtek:
changed the path of Roundcube config file

Did the installation today and found this "issue".


andryyy:
Merci

PR #185, opened 2020-01-01
Update migration guide to use Admonition plugin and some small fixes

MAGICCC:
Migration guide looks like this now:
image


andryyy:
Merci

PR #186, opened 2020-01-30
Update rewriterule for Apache to get Letsencrypt working

Mikesch-mp:
The rewrite rule only checked for non https and then rewrote every request to https. Letsencrypt checks use http only so we have to let them stay on http and not rewrite them to https.


andryyy:
Hm, im using this on many setups as is. 

It should follow the redirect just fine. :)

Mikesch-mp:
It follows, but not let thru letsencrypt for mailcow on http.

PR #189, opened 2020-02-14
update NextCloud oauth link

pgollor:
In my Nextcloud 18 it only works with `https://cloud.example.com/apps/sociallogin/custom_oauth2/Mailcow` instead of `https://cloud.example.com/index.php/apps/sociallogin/custom_oauth2/Mailcow`.

### DIFF
--- a/docs/third_party-nextcloud.md
+++ b/docs/third_party-nextcloud.md
@@ -4,7 +4,7 @@ In the following, we will only assume that you have already set up NextCloud at 
 To set up authentication via mailcow, you can use OAuth2 as described below. 
 
 1. Log into mailcow as administrator.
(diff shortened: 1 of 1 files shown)


mkuron:
Actually, this is not related to the NextCloud version, but only depends on how your rewrite rules are configured. You don't need it when you use Mailcow's nginx template (https://github.com/mailcow/mailcow-dockerized/blob/cd88165282da09f2d3dd393201b6a2489ed5f542/data/assets/nextcloud/nextcloud.conf#L67), but you do need it when using e.g. when running the NextCloud docker image (https://hub.docker.com/_/nextcloud).

So please change your pull request to something like

> Some installations, including ones set up using the [helper script](https://raw.githubusercontent.com/mailcow/mailcow-dockerized/master/helper-scripts/nextcloud.sh), need to remove `index.php/` from the URL to get a successful redirect.

mkuron:
The merged PR still contained the incorrect NC 18 reference (for NextCloud 18 without `index.php`), please remove.

andryyy:
Ooops. :( Fixing it now.

PR #191, opened 2020-02-16
Added missing port in netstat command

DeatPlayer:
In the netstat command for checking the standard-ports the port for "Dovecot ManageSieve" was missing


andryyy:
Thanks

PR #193, opened 2020-02-29
Saving other hours of frustration with a few words per https://github…

dgtlmoon:
per https://github.com/mailcow/mailcow-dockerized/issues/3373

### DIFF
--- a/docs/i_u_m_install.md
+++ b/docs/i_u_m_install.md
@@ -39,9 +39,14 @@ cd mailcow-dockerized
 ```
 
 **4\.** Change configuration if you want or need to.
+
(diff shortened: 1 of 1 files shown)


dgtlmoon:
@andryyy Locked me out of the parent issue so I could not respond, that's really not cool.

The build script uses a default network which is probably in use by anyone who has other docker projects in place, That's really not very cool behaviour that you would just block an issue and not hear out someone's opinion on it.

Who is the project lead here?

dgtlmoon:
https://github.com/mailcow/mailcow-dockerized/issues/3373

PR #197, opened 2020-03-10
Create BMX from scratch

djdomi:
i created a simple bmx manual


djdomi:
Please check if that's working so far, only missing the part to stop and start the container, as i don't know the names yet

mkuron:
There are a number of typos and formatting errors that need to be fixed. Furthermore, this should definitely not be on the main install guide as the average user does not need a backup MX. Backup MXs have some peculiarities (such as the spam filter not being trainable and needing to create all user accounts on both primary and backup MX) and thus require careful planning.

MAGICCC:
Also a small explanation for who and what a BackupMX is would be neat

djdomi:
> There are a number of typos and formatting errors that need to be fixed. Furthermore, this should definitely not be on the main install guide as the average user does not need a backup MX. Backup MXs have some peculiarities (such as the spam filter not being trainable and needing to create all user accounts on both primary and backup MX) and thus require careful planning.

This SEtup does not care about the re-creation of all, its kept as simple as possible

djdomi:
> Also a small explanation for who and what a BackupMX is would be neat

I agree to you, as told i kept it as simple as possible

andryyy:
Something is wrong with this PR. It is grammatically and stylistically incorrect, too. :/

mkuron:
> This SEtup does not care about the re-creation of all

It should as it's a bad idea to not check recipients. Otherwise invalid addresses can cause delayed bounces, which may become backscatter spam.

djdomi:
> Something is wrong with this PR. It is grammatically and stylistically incorrect, too. :/
@diekuh i updated that

mkuron:
The pull request currently just deletes the entire content of the file, which is not what you had intended. Please fix that and address the comments above if you haven't done so already.

djdomi:
ii gona close and re-open this, as i dont know how could i restore this file

PR #201, opened 2020-03-16
Updated Prerequisites

Gaming4LifeDE:
The file was edited to remove the outdated (and in some cases removed) `netstat` command and replace it with `ss`


andryyy:
Added both.

Gaming4LifeDE:
Maybe we should add a note that the `ss` command is the only one that'll work on newer distros.

PR #203, opened 2020-03-16
Added instructions for the Nine E-Mail app

Gaming4LifeDE:
I did this because this is the only Android app which allows you to see calendar categories which can be synced with the server.
I added a disclaimer saying that mailcow is not affiliated with the company which produces the app and explaining that it's found to work properly and that's all. Feel free to edit this in any way you see fit.


andryyy:
Should work with autodiscover. And it also exists for iOS. :)

Gaming4LifeDE:
Didn't work for me at least.

Gaming4LifeDE:
Works with other clients though but just imap/smtp and caldav

PR #205, opened 2020-03-18
disable network for certdumper

vidschofelix:
prevent certdumper from creating his own network and messing around with the backup/restore-script. Certdumper doesn't need network at all


andryyy:
Merci :)

PR #210, opened 2020-04-20
Add info on importing from GMail

philipgo:


andryyy:
Use `--skipcrossduplicates` or `--gmail1`. Iirc `--gmail1` includes skipcrossduplicates. :)

PR #212, opened 2020-04-23
modernized nginx reverse proxy example

707Alex707:
Updated Nginx reverse proxy example
- Disabled TLSv1.0 and TLSv1.1
- Enabled TLSv1.3 (TLSv1.2 was already enabled)
- Enabled http2
- Updated supported ssl_ciphers (with Mozilla recommended ciphers)  

Additional formatting fixes

### DIFF
--- a/docs/firststeps-ip_bindings.md
+++ b/docs/firststeps-ip_bindings.md
@@ -1,5 +1,5 @@
-   !!! warning
-   Changing the binding does not affect source NAT. See [SNAT](https://mailcow.github.io/mailcow-dockerized-docs/firststeps-snat/) for required steps.
+!!! warning
+    Changing the binding does not affect source NAT. See [SNAT](https://mailcow.github.io/mailcow-dockerized-docs/firststeps-snat/) for required steps.
 
 ## IPv4 binding
 
--- a/docs/firststeps-rp.md
+++ b/docs/firststeps-rp.md
@@ -111,8 +111,8 @@ server {
   ssl_session_cache shared:SSL:50m;
   ssl_session_tickets off;
 
-  ssl_protocols TLSv1.2 TLSv1.3;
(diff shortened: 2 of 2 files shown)


andryyy:
What about systems without TLS 1.3 support?

Please comment it out to make it optional with a small hint. :) Same with the ciphers. A link to the Mozilla tool is enough.

707Alex707:
Fair point on the TLS 1.3 support. 

Ciphers though, should probability still be updated. As the current "strong" ciphers is a subset of the "HIGH:!aNULL:!MD5" ciphers. So server side compatibility should be maintained, provided TLS 1.3 ciphers aren't included. 

@andryyy Let me know what you think.

andryyy:
Sounds good

707Alex707:
Rather then specifying a hard list of ciphers, I just disabled those using SHA1 and RSA for key exchange as they are obsolete. Ciphers using CBC are still enabled even though they are obsolete. There isn't a !CBC option so I would have to specify cipher by cipher to disable, which may break compatibility if the cipher doesn't exist in their openssl version. 

If ciphers were hard set, some newer ciphers such as ECDHE-ECDSA-CHACHA20-POLY1305
ECDHE-RSA-CHACHA20-POLY1305 would have to not be included as they are recent additions to openssl. With the setting of "HIGH:!aNULL:!MD5:!SHA1:!kRSA" new ciphers will be supported automatically if their system supports them. While there may be a few "weak" ciphers, clients in theory should pick "strong" ones. 

Side Note, Mozilla's ssl tool mislead me, changing the openssl version only toggles if TLSv1.3 is enabled or not along with the corresponding ciphers. It doesn't change the list of ciphers if the cipher isn't supported in a particular version of openssl which is kind of odd. 


> Please comment it out to make it optional with a small hint. :) Same with the ciphers. A link to the Mozilla tool is enough.

This is a bit confusing, do you want them commented out or just a link to the Mozilla tool? There is both write now. If you want that changed, I'll happily change it.

andryyy:
Thank you for your explanation!

I thought about a link like...

```
# blabla
# https://ssl-config.mozilla.org/
```

PR #213, opened 2020-04-23
replaced enable command in documentation

lehuizi:


andryyy:
This will not work on older systemd integrations like possibly with C7

PR #215, opened 2020-05-06
Default relay behaviour for RFC1918

Programmierus:
Adds some warnings about default mailcow behavior as a relay for private networks. In some cases, this might be dangerous and users must be aware.


andryyy:
We talken on Telegram about it. :)

PR #216, opened 2020-05-06
Mention lz4 compression in u_e-dovecot-mail-crypt.md #2754, #2337

jrenggli:
I found two issues (#2754, #2337) discribing problems with unreadable files after decryption. This is caused by the compression with lz4.
By mentioning this in the documentation it could be easier for other users.


andryyy:
Nice! Works great!

PR #218, opened 2020-06-01
Make traefik use right net address of nginx

leifnel:
Unless specified, traefik might try use the ip of nginx on the mailcowdockerized_mailcow-network, which is not reachable.


andryyy:
merciiii :)

PR #221, opened 2020-07-02
Update third_party-roundcube.md: managesieve

johanvandyck:
The current doc gives "cannot connect to db" in Roundcube 1.4.6, managesieve. The manage sieve config should be placed in the plugin itself and not in the general config file of Roundcube.


andryyy:
Thank you, looks good. :)

PR #222, opened 2020-08-12
SSLEngine must be active

markusmarkusz:
In order for SSL to work correctly with apache2, SSLEngine must be active for the VirtualHost.
Otherwise the SSL certificate will be ignored.

### DIFF
--- a/docs/firststeps-rp.md
+++ b/docs/firststeps-rp.md
@@ -43,7 +43,7 @@ Let's Encrypt will follow our rewrite, certificate requests in mailcow will work
 
 **Take care of highlighted lines.**
 
-``` apache hl_lines="2 10 11 17 22 23 24 25 30 31 32"
+``` apache hl_lines="2 10 11 17 22 23 24 25 30 31"
 <VirtualHost *:80>
   ServerName CHANGE_TO_MAILCOW_HOSTNAME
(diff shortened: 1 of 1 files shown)


andryyy:
I think this depends on the Apache version. I use this exact config for many rps. :/

markusmarkusz:
Do you have SSLEngine for Apache directly active? There seems to be a way to do this.

Otherwise we should perhaps then add a comment that it is sometimes necessary.

I had problems because apache always took the SSL certificate from another VirtualHost. After I set SSLEngine to on in this VirtualHost, the problem no longer existed.

Since the bug seemed more general to me, I just did this PR.

andryyy:
Iirc most distros enable the engine when you load/install the module. Not sure though. Which distro did you test on?

We should definitely add it as comment. That's a good idea. :)

markusmarkusz:
I'm using Debian 10.

I'll adjust the PR later.

andryyy:
Oh, that's interesting. Did you enable SSL via a2enmod ssl?

markusmarkusz:
Completely forgotten. Is that enough?

And yes, I did.

markusmarkusz:
why closed? (maybe @andryyy?)

PR #224, opened 2020-09-19
Update i_u_m_install.md

OKNoah:
Clarifies getting latest docker compose.


andryyy:
Hi, could you please change "should" to "will"? :)

Getting it via the command shown above makes sure we can skip incompatible or broken versions. 

I do normally just pass the latest version available, in some rare cases (it happened once) we can control the pushed version.

PR #225, opened 2020-09-20
Update third_party-gogs.md

bimre:
Change proxy_pass http://gogs:3000/; port to 4000


andryyy:
Why proxy pass to the SSH port?

PR #227, opened 2020-09-24
Adds consistency to hard-to-notice example code

OKNoah:


andryyy:
Hm, I don't know. It is pretty obvious. :(

OKNoah:
@andryyy Then explain. If all other instances are coloured yellow and have clear instructions to change it then.. it’s clear this one should not be changed, right?

MAGICCC:
Then we would need to edit basicly the whole docs.
Also why did you change the domain to a .com tld on this line and not everwhere else?

OKNoah:
It’s great that others found it obvious. But would anyone be upset if we made the grey-on-grey placeholders more visible.

andryyy:
We can hopefully assume people who do run their own mail servers do not copy a line with a DMARC TXT they A) don't understand and B) includes an @example.org address. 

mailcow makes mail easy, yes, but that's too much imo.

OKNoah:
Remaining in the point of visibility , can you tell me how the TLD matters?

My anecdote: I’ve installed Mailcow 3 times and never noticed that grey’d-out placeholder email.  Today I finally went to ask on IRC and didn’t get an answer. 

There’s also the issue of “mailauth-reports” being in the email address, which seems more specific that “foo” or “john appleseed”, further suggesting it’s in while or in part not a placeholder. 

There’s actually anotherglaring issue in the page and I feel it should be rewritten. E.g. one section on this page leaves a value blank and tells the reader to “learned the answer by studying this linked sites.

Said links go to the home page, not the relevant reference material. Within the sites are 14 sections of documentation 

When I first tried to run an email server long before Docker, it meant following a massive guide and one typo meant starting over. 

Docker has removed a lot of steps and concerns but it still takes docs, and if we opposing iterative change because ... you know, then self hosting will continue  to take losses. 

No offence to the hard workers, but I may have plans for a fork that brings back the happier cow.

PR #230, opened 2020-10-02
Update prerequisite-dns.md

OKNoah:
Trying this kinda thing on Linode, not sure if working just yet THB

### DIFF
--- a/docs/prerequisite-dns.md
+++ b/docs/prerequisite-dns.md
@@ -64,6 +64,21 @@ _carddavs._tcp      IN TXT     "path=/SOGo/dav/"
 _caldavs._tcp       IN SRV     0 1 443   mail.example.org.
 _caldavs._tcp       IN TXT     "path=/SOGo/dav/"
 ```
+## Linode-CLI
+
+Since this is a lot to enter into many DNS interfaces, you might try a CLI, such as Linode supplies. Each dommand looks somethig like this:
+
+```sh
+linode-cli domains records-create \
+ --service _imaps._tcp \
+ --type SRV \
+ --weight 0 \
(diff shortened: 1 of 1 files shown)


shiz0:
Hm.. is it working, or not? Did you test it?
Also, while I totally get your point that adding entries via a CLI would be nicer than using a web interface like most DNS providers provide (including my own), I think that provider specific instructions may be beyond the scope of this documentation.
The community board may be a better place for such examples/howtos? 
That's just my personal thinking, though.

andryyy:
I also think this is not the right place. Plus I don't really want to advertise Linode tbh.

I think if we provide examples provider based, Servercow would be _kind of fair_, don't you think? :) I even try to avoid _that_.

MAGICCC:
I agree with @shiz0 + @andryyy 
Either we add a section for (common) DNS provider on the docs or the board, but only one is not good.
Even though it's missing a how to install this cli

OKNoah:
> I also think this is not the right place. Plus I don't really want to advertise Linode tbh.

Curious what your beef with Linode is. I wrote this as a start to having as many examples as possible, 1 for each host that has a DNS API and someone who can submit a command for it.

We may also choose to use a tabbed view, like with code example in API docs, until we run out of space there. Then we make a separate page or organize another way.

But I think it's sometimes foolish to not just start with one and keep it simple. It's a bike in need of a shed, the bike is useful immediately without a shed.

So this saves, I dunno, 5% of people a lot of effort. Then whoever adds servercow can organize it with a tab view, etc and help some more people, and so on. It's not like an example, it's specific to each host with DNS. I just wanted to start with one and see it change as more contributed their own.

PR #231, opened 2020-10-03
Remove EoL updating method

MAGICCC:


ntimo:
Thanks, I think removing it is a good idea 👍 since it can potentially break a lot of setups

andryyy:
Yes

PR #232, opened 2020-10-05
Add another way of connecting Cal/CardDAV to TB

Braintelligence:


andryyy:
Merci. A precise step by step documentation would be great though. :) We cannot use "should be there" in a documentation. 

If you are able to, some screenshots would be amazing. :)

Braintelligence:
@andryyy How's this? 😁

PR #234, opened 2020-10-10
Add Exchange Hybrid Setup

JUVOJustin:
Add Docs about how to use a mailcow as on premise mailserver in combination with office 365 / Exchange

### DIFF
--- a/docs/u_e-exchange-onprem.md
+++ b/docs/u_e-exchange-onprem.md
@@ -5,7 +5,7 @@ This setup becomes very handy if you have enabled the [Office 365 security defau
 
 
 ## Requirements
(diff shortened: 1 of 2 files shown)


andryyy:
Hi Justin,

Thank you!

Perhaps we should add the MS servers as forwarding host and make clear that filtering now completely relies on MS (ieks, not recommended).

Inbound filtering mailcow side is broken so is neural learning etc. - we need to completely disable these components.

JUVOJustin:
Hi @andryyy , 
you´re absolutely right. I added a step to add the forwading host and an notification box about the broken inbound filtering.

I didn´t found any documentation about forwardong hosts that i could link to. Is there any?

andryyy:
There is none yet. :( But it is enough to point to /admin and "Forwarding hosts". :-)

JUVOJustin:
Alright i added the path to the forwarding hosts menu. 

I think this documentation is in a "ready to be merged" state, isn´t it? :D

andryyy:
Amazing!

Ready to merge for you?

JUVOJustin:
Yup. Go for it 👍

andryyy:
Very cool, thank you, again.

PR #235, opened 2020-10-13
Upload Nextcloud document

heavygale:
- This change splits the Nextcloud dokument in two sections. One describing how to use the helper script and the second for how to configure Nextcloud to use mailcow for authentication.
- I also added a hint to mention the option for resetting a Nextcloud password as added in https://github.com/mailcow/mailcow-dockerized/pull/3806
- And I hopefully fixed the wrongly listed (ordered list or no list instead of unordered list) bullet points by removing the indentation and adding blank lines: 
grafik

### DIFF
--- a/docs/third_party-nextcloud.md
+++ b/docs/third_party-nextcloud.md
@@ -1,12 +1,11 @@
 
-## Manage the Nextcloud using the helper script
+## Manage Nextcloud using the helper script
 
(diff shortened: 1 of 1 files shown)


andryyy:
Merci!

heavygale:
At least at partly success,... but 6 to 9 are still listed with numbers instead of bullet points, and the numbers start over with 1 after the info-box. 🤨 How do those lists work?

grafik

PR #236, opened 2020-10-14
Add Note About DNSBL

seniorm0ment:
added note reminding to check postfix logs if mail is being denied with an ipv6 error. After long troubleshooting it would have been nice to have remembered this.


dragoangel:
I think this is good note

PR #237, opened 2020-10-14
Added headers, and new resources

grravity:
- Added headers to the `# Advanced DNS Configuration` table,  the `# DKIM, SPF and DMARC` section, and #` DMARC` section, for improved readability, as well as to match the top portion which has headers.
- Added [MultiRBL.valli.org](http://multirbl.valli.org/) as a resource for testing DNSBL, RBL, and FCrDNS under `# TEsting`
- Added various resources for DMARC analysis services.
- Moved around, and modified the `# Misc` section to provide more detailed DMARC analysis information.
- Added a reference for DMARC at the top.

Edit: I apologize for the commit spam.

### DIFF
--- a/docs/prerequisite-dns.md
+++ b/docs/prerequisite-dns.md
@@ -89,21 +89,15 @@ If you are interested in statistics, you can additionally register with some of 
 
 **NOTE:** It is worth considering that if you request DMARC statistic reports to your Mailcow server, if there are issues with that domain you may not get accurate results. You can consider using an alternative email domain for recieving DMARC reports.
 
+It is worth mentioning, that the following suggestions are not a comprehensive list of all services and tools avaialble, but only a small few of the many choices.
+
 - [Postmaster Tool](https://gmail.com/postmaster)
 - [parsedmarc](https://github.com/domainaware/parsedmarc) (self-hosted)
 - [Fraudmarc](https://fraudmarc.com/)
 - [Postmark](https://dmarc.postmarkapp.com)
 - [Dmarcian](https://dmarcian.com/)
 
-The services should provide you with a TXT record, which will give you details about spam-classified mails by your domain.
-Using Google Postmaster as an example:
+**NOTE:** The services may provide you with a TXT record, which you would insert into your DNS records as the provider specifies. This record will give you details about spam-classified mails by your domain. However, please ensure to read the providers documentation from the service you choose, as this process may vary and not all providers may use a TXT record.
 
-```
-# Name              Type       Value
-
-@                   IN TXT     "google-site-verification=..."
-```
-
-However, ensure to read documentation from the service you choose, as this process may vary.
-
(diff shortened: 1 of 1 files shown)


andryyy:
I don't think I want to list this as prerequisite. And furthermore not with some providers listed and others not. Or the Google verification TXT.

You can list that somewhere else, but it is not a prerequisite, should not contain providers without a hint like "there are other providers blabla" and not with a Google TXT when a TXT may not be required by other providers. :/

grravity:
> I don't think I want to list this as prerequisite. 

What exactly do you not want? Just the modification to the DMARC section?

Or everything, including the headers added to the tables? In my personal opinion, having the "Priority, Weight, Port" headers above the advanced DNS configuration would've saved me a lot of hassle when 

___________

> And furthermore not with some providers listed and others not. 

The original reason I added a few more providers, is to give clarity that more providers do exist, as it is understandable to me that not everyone would wish to use Google. I especially think that the self-hosted one may have been in interest of the users.

> You can list that somewhere else, but it is not a prerequisite

I don't entirely disagree, however I think that the DMARC statistics piece that is already there should be moved to it's own sub page in that case. But I also personally don't see a major issue with it being where it is, as if people are already inputting records, this is where they would want to know about it. So at very least, a link to a sub-page with DMARC statistics should be made there imo.

> Should not contain providers without a hint like "there are other providers blabla" 

`However, ensure to read documentation from the service you choose, as this process may vary.`
I did provide that immediately after the Google text, I figured this would have been sufficient. I also stated that the Google TXT was an example, and that services should provide you with a TXT file (although maybe should could be changed to may). 

` you can additionally register with some of the many below DMARC statistic services, or self-host your own.` that was my attempt at a hint in suggesting other providers exist besides the ones listed.

> and not with a Google TXT when a TXT may not be required by other providers.

I provided it as an example and because I didn't feel like it should have been taken away from the original statement.
I did however alrea

I figured moving things around, and wording it like this, in attempt to be more specific, and suggesting as I did that the process may vary but still wanted to give an example so I kept the google TXT. I also seperated the section so it's it's own sub-section under Misc so it's a bit more seperated. Again, up to you just figured I'd explain my thought process.

dragoangel:
Agree with Andre that Google Postmaster better to be excluded.
There many postmaster and FBL programs, it looks strange to add this and not add other.
But at same time they *mostly* needed for **Bulk** mail senders.

And if speak about exact Google tool: it will show you some data only in case you have approximately more then 300 email/day for domain. Most people will try to configure it but will not see any data.

grravity:
I'm just confused on what needs to be changed or removed from my commit, do you want me to just remove the codeblock for Google Postmaster, and make it more clear that other services exist? (Which I already did,  However, ensure to read documentation from the service you choose, as this process may vary. , but it may not be entirely clear?)

PR #239, opened 2020-10-24
Note about 2GB+Swap, host port blocks & minor tweaks

grravity:
- Added a secondary resource (slightly more detailed) for docker iptables setup which has been referenced and used numerous times by a few people in the chatroom.
- Added note about the possibility of hosting on only 2 GiB + Swap, as this has been done by multiple users (for only hosting one person), however still not ideal and recommended. I added this as users may wish to only host for themselves, and many VPS plans offer 2GB ram plans, and this can save users money if they know that it is possible to do it on less than 4 GiB. 
- Tweaked the note about ClamAV and Solr.
- Added a section about some server providers and ISP's blocking ports.


andryyy:
No, please don't change requirements of our stack by yourself. 2 GB is not enough. 4 isn't even enough with 20 workers.

PR #240, opened 2020-10-29
Update prerequisite-dns.md

mattronix:
Added an email test that allows you to test both sending and standard spam / DKIM Checks


andryyy:
Thanks

PR #241, opened 2020-11-03
Add dovecot missing ssl files as common problem

balping:
See https://github.com/mailcow/mailcow-dockerized/issues/2672

### DIFF
--- a/docs/debug-common_problems.md
+++ b/docs/debug-common_problems.md
@@ -72,7 +72,22 @@ dism /online /Enable-Feature /FeatureName:TelnetClient
 
 Docker containers use the Docker hosts inotify limits. Setting them on your Docker host will pass them to the container.
 
+## Dovecot keeps restarting (see [#2672](https://github.com/mailcow/mailcow-dockerized/issues/2672))
 
+Check that you have the following files in `data/assets/ssl`:
+
+
+```
+-r--r--r-- 1 root root 3554 Nov  3 17:07 cert.pem
+-rw-r--r-- 1 root root  769 Nov  3 18:24 dhparams.pem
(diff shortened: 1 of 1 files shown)


andryyy:
These file are there by default. It is actually not a common problem. :)

Furthermore your file permissions are different from the original files. That may confuse people. 

After bootstrapping acme-mailcow one can have even more files and folders there. 

I think that's not a common problem and people should follow and understand the docs when exchanging the certificates.

balping:
> These file are there by default. It is actually not a common problem. :)

Well, I ran into the same issue twice while installing mailcow on two servers, separated roughly by a year and I just thought I would spare myself next time (and others) from this hour of painful debugging. (It *is* painful as the container restarts every couple of seconds and therefore you can't really read any logs.)  But I can see  `generate_config.sh` should copy  `dhparams.pem` from  `data/assets/ssl-example` to `data/assets/ssl`, so I don't know what I did wrong, maybe I deleted it accidentally. There's also a number of people who encountered it in the thread linked above.

> Furthermore your file permissions are different from the original files. That may confuse people.

I don't mind changing the permissions to the default ones, I just copied the output of `ll`

> After bootstrapping acme-mailcow one can have even more files and folders there.

True, maybe I should add something along the lines *"when using custom certificates"*

balping:
@andryyy I incorporated some of your notes. If you think this is unnecessary, feel free to close it.

Blackclaws:
This issue is common if you don't run generate-config but instead copy an existing mailcow.conf over from a different location and customize it manually. I just ran into this as well because I had to recheckout the data folder but I already had the config.

PR #242, opened 2020-11-09
Mailman integration

pgollor:
Add mailman to third party apps with scripts from https://github.com/pgollor/mailman-mailcow-integration.
Someone have to be test the mailman integration before adding to the official documentation. Thanks for that.
The help file is temporary hosted [here](https://gist.github.com/pgollor/204258f4ca4d10aed5706eae791f83b3).

### DIFF
--- a/docs/third_party-mailman.md
+++ b/docs/third_party-mailman.md
@@ -1,0 +1,88 @@
+Mailman is a system to manage email lists with a web UI.
+The following guide uses the docker-compose implementation from [maxking](https://github.com/maxking/docker-mailman) depends on postgres database with [mailman-core](https://github.com/maxking/docker-mailman/tree/master/core) and [mailman-web](https://github.com/maxking/docker-mailman/tree/master/web).
+
+!!! info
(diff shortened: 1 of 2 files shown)


andryyy:
Interesting. :)

I will try it hopefully soon. 

What do you need to test?

If you pass the mailman domains to the php container, we could list them as "in use by mailman" or something like that.

pgollor:
> Interesting. :)
> 
> I will try it hopefully soon.

Thank you.

> What do you need to test?

Time ;-) because i have to two little children and a lots of interesting software and hardware project ideas.
I think the best way is to install a fresh mailcow installation for a subdomain, use another subdomain to feed my [installation script](https://github.com/pgollor/mailman-mailcow-integration/blob/master/mailman-install.sh) for Mailman and create a test list.
The integration works for me over years but I wrote quick and dirty the documentation and the installation script. **Maybe I forgot to add or document something.**

pgollor:
Apart from me, no one else seems to need a mailman integration in mailcow.

PR #244, opened 2020-11-20
added new section to the Dovecot expunge page explaining a way to automatize with Ofelia

grimneko:
Since I run it with Ofelia instead of the hosts cron I wrote a little addition for the documentation to enable everyone to also do it this way. Ofelia is a low-memory footprint docker job scheduler running on docker which makes it easy to run jobs in docker containers utilizing a cron style time string.


andryyy:
Merci. :)

Can you replace "Make it bla" by `Schedule job`? I know that's not your change, but it sounds better.

Could you please add a full docker-compose.override.yml including the service header? That'd be cool.

andryyy:
Thank youuu

PR #245, opened 2020-11-30
Update firststeps-rp.md

BlackWolfDEsign:
it looks like the old version no longer works. no more redirection ... changed, tested and redirects now again


andryyy:
Hm, I don't think it is broken. Which version of Apache2 did you try?

We can merge it anyway, I don't see a problem with !=on or =off. :)

PR #248, opened 2020-12-23
Order DNS records alphabetically and fix capital m (content not altered)

Bojan023:
original record with capital M: _carddavs._tcp      IN SRV     0        1      443      **M**ail.example.org.


andryyy:
Thank you!

PR #251, opened 2021-01-10
Feature/jwilder nginx

maxcodeshere:
Added a working config for the jwilder/nginx and & letsencrypt-nginx-proxy-companion setup that utilizes overrides. It should hopefully reduce the number of opened issues referring to this particular setup.


andryyy:
How are Dovecot and Postfix informed and restarted?

PR #255, opened 2021-01-30
Add info about SOGo custom themes and favicon

dragoangel:


andryyy:
Thanks. The "revert theme to default" guide is a bit different from the original file, but if it works, it works.

PR #257, opened 2021-02-07
Internal subnet

Op3rat0r:
Adding the internal subnet to the nc configuration. Helped to access the nc instance via my revers proxy. It was accesible from outside, just not from inside.

### DIFF
--- a/docs/third_party-nextcloud.md
+++ b/docs/third_party-nextcloud.md
@@ -84,13 +84,13 @@ The Nextcloud instance can be updated easily with the web update mechanism. In t
 It shows which commands have to be executed, these have to be placed in the php-fpm-mailcow container.
 
 As an an example run the following command to add the missing indices.
-`docker exec -it -u www-data $ (docker ps -f name = php-fpm-mailcow -q) bash -c "php / web / nextcloud / occ db: add-missing-indices"`
+`docker exec -it -u www-data $ (docker ps -f name = php-fpm-mailcow -q) bash -c "php /web/nextcloud/occ db: add-missing-indices"`
 
 ---
 
 ## Debugging & Troubleshooting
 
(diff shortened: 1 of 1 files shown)


Op3rat0r:
It's ok to just put the external domain into the trusted_proxy list.

andryyy:
The trusted proxy is set by the helper script, isn't it?

Op3rat0r:
In the array are two entries.
0 => 'fd4d:6169:6c63:6f77::/64',
1 => '172.22.1.0/24',

It works fine when I called the url from outside my network. But didn't work when you call the url from the internal nw. After that changes it works.

andryyy:
Ah, okay, misunderstood it. :)

docs/third_party-nextcloud.md
```diff
@@ -75,3 +75,31 @@ If you have previously used Nextcloud without mailcow authentication, but with t
...
+
+---
+
+## Update
+
+The Nextcloud instance can be updated easily with the web update mechanism. In the case of larger updates, there may be further changes to be made after the update. After the Nextcloud instance has been checked, problems are shown. This can be e.g. missing indices in the DB or similar.
+It shows which commands have to be executed, these have to be placed in the php-fpm-mailcow container.
+
+As an an example run the following command to add the missing indices.
+`docker exec -it -u www-data $ (docker ps -f name = php-fpm-mailcow -q) bash -c "php / web / nextcloud / occ db: add-missing-indices"`
```
andryyy, line 87: Are there spaces between the path? Like "php / web / nextcloud"?

Op3rat0r: Oh, no there are no spaces. It's a pathe "/web/nextcloud/occ"

docker exec -it -u www-data $(docker ps -f name=php-fpm-mailcow -q) bash -c "php /web/nextcloud/occ db:add-missing-indices"

andryyy:
I removed the spaces. :) Thank you!

Op3rat0r:
You are welcome! I am happy that I could contribute something to the cow ;)

PR #258, opened 2021-02-07
Updated copyright year

guiguir68:
and HNY!


MAGICCC:
hah bit late but thanks!

PR #259, opened 2021-02-08
Extra space character in dovecot expunge docker-compose.override

Rijul-Ahuja:
Remove the extra space character at the start of the services line in the ofelia docker-compose.override.yml. PR prevents the below error.
```
ERROR: yaml.parser.ParserError: while parsing a block mapping
  in "./docker-compose.override.yml", line 1, column 1
expected <block end>, but found '<block mapping start>'
  in "./docker-compose.override.yml", line 2, column 2
```


andryyy:
Thanks :)

PR #228, opened 2020-09-24
Add DMARC Reporting Docs

dragoangel:

PR #260, opened 2021-02-08
dmarc reports example configuration

Op3rat0r:
As the question of how to activate the dmarc reports on the cow has repeatedly arisen, here is a short example of the documentation.


andryyy:
There should be an open PR about this which needs spell checking. Can you work with that?

Op3rat0r:
Ah sry didn't see. I'll habe a look at that.

MAGICCC:
For reference https://github.com/mailcow/mailcow-dockerized-docs/pull/228

PR #262, opened 2021-03-03
Update build-thunderbird-plugins.sh to make it compatible witth thunderbird v78

etlam:
SOGo restructured their connector so the build-thunderbird-plugins.sh script does not work for Thunderbird 78+.  
On the other hand a release in the SOGo connector repository is missing, so we have to use the master branch:
https://github.com/inverse-inc/sogo-connector/tree/master

The update script is going to create a customized version of SOGo connector, that works with Thunderbird v78+.
We are working with the generated SOGo connector in production and for example HU-Berlin used the master-branch too, to be able to support Thunderbird v78+:
https://kal.hu-berlin.de/plugins/

### DIFF
--- a/docs/download/build-thunderbird-plugins.sh
+++ b/docs/download/build-thunderbird-plugins.sh
@@ -10,27 +10,28 @@ fi
 
 cd $(dirname $0)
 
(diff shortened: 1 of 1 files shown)


mkuron:
Thanks! I briefly tried fixing it myself a while ago but ended up switching to TbSync, which seems more reliable and significantly better supported these days.

PR #264, opened 2021-03-31
Update third_party-gitea.md

fabreg:
Added a step needed if you use http to https redirect

### DIFF
--- a/docs/third_party-gitea.md
+++ b/docs/third_party-gitea.md
@@ -33,11 +33,13 @@ GITEA_SSH_PORT=127.0.0.1:4000
 
 5\. Run `docker-compose up -d` to bring up the gitea container and run `docker-compose restart nginx-mailcow` afterwards.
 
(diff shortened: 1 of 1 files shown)


andryyy:
Hmm, the redirect does redirect to https anyway, doesn't it?

We are referencing https as root_url in the app.ini, so we can probably just change the docs to https-only and replace the http parts, what do you think?

fabreg:
Ciao @andryyy, yes it redirect to https - I'm talking about this redirect: https://mailcow.github.io/mailcow-dockerized-docs/u_e-80_to_443/
With this enable, if you try to browser http://mx.xxx.com/gitea you will see the broken page without CSS loaded (classic problem of http to https redirect)
IMHO, it's better to give the user the possibility of choice: if you use http follow the steps, if you use https perform this additional step.
I know in 2021 it's kinda weird to not use https but maybe be someone, for an internal environment, prefers to keep on http.
However, last word is yours :)

andryyy:
I see. For me it's fine. We cover all scenarios then. :) Thanks!

Issue #265, opened 2021-04-01
SOGo theming docs


vizion8-dan:
Hi, thanks a lot for bringing SOGo-themeing back!

I don't know if its missing in [documentation](https://mailcow.github.io/mailcow-dockerized-docs/u_e-sogo/) or its just me not understanding it right or my configuration - but theming works for me only if I manually add `SOGoUIxDebugEnabled = YES;` to `data/conf/sogo/sogo.conf`.  (Tested on two different servers).

andryyy:
Check https://www.sogo.nu/files/docs/SOGoDevelopersGuide.html

:)

andryyy:
This should be part of the docs. Do you mind creating a PR?

vizion8-dan:
Sure, I'll create a PR over the weekend.

andryyy:
Thank you!!

------ Originalnachricht ------
Von: "Adriano" ***@***.***>
An: "mailcow/mailcow-dockerized-docs" 
***@***.***>
Cc: "André Peters" ***@***.***>; "State change" 
***@***.***>
Gesendet: 01.04.2021 21:40:53
Betreff: Re: [mailcow/mailcow-dockerized-docs] SOGo theming docs (#265)

PR #266, opened 2021-04-01
Add reference to "SOGoUIxDebugEnabled = YES"

vizion8-dan:
This is for #265 

After testing SOGo's theming features, following the provided [documentation](https://mailcow.github.io/mailcow-dockerized-docs/u_e-sogo/) I figured out that an additional `SOGoUIxDebugEnabled = YES;` was needed in `data/conf/sogo/sogo.conf` to make it actually work as expected (= as in works for me). As I am currently not aware of any side effects using `SOGoUIxDebugEnabled = YES;` in production - this should be checked by someone who is more into the SOGo....

The [SOGo Developer’s Guide](https://www.sogo.nu/files/docs/SOGoDevelopersGuide.html) mentions using `SOGoUIxDebugEnabled = YES;` "temporarily" and updating some .css files with generated style information. I don't know how this would fit into the mailcow system or if its just okay to go with `SOGoUIxDebugEnabled = YES;` .

Maybe the docs should also link/reference the part of the SOGo Developer’s Guide?

### DIFF
--- a/docs/u_e-sogo.md
+++ b/docs/u_e-sogo.md
@@ -1,10 +1,15 @@
 
 SOGo is used for accessing your mails via a webbrowser, adding and sharing your contacts or calendars. For a more in-depth documentation on SOGo please visit its [own documentation](http://wiki.sogo.nu/).
 
-## Change theme
+## Apply custom SOGo theme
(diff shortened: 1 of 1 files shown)


andryyy:
Thank you!!

PR #267, opened 2021-04-16
Third-party: Borgmatic Backup

kaechele:
This guide describes setting up automatic recurring backups using
borgmatic.
This is a good alternative to juggling tar archives.

Signed-off-by: Felix Kaechele <felix@kaechele.ca>

### DIFF
--- a/docs/third_party-borgmatic.md
+++ b/docs/third_party-borgmatic.md
@@ -1,0 +1,247 @@
+# Borgmatic Backup
+
+## Introduction
+
+Borgmatic is a great way to run backups on your Mailcow setup as it securely encrypts your data and is extremely easy to
+set up.
+
+Due to it's deduplication capabilities you can store a great number of backups without wasting large amounts of disk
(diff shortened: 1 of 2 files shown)


andryyy:
Thank you very much for this! :) I try to check it tomorrrow and report back.

kaechele:
FYI: I have actually tested this guide on a clean mailcow install (both backup and restore) and it worked for me.
So I hope this is useful for people.

I run it on my instance pretty much in this configuration and with the large amount of snapshots available I can always so go back in time without messing up things too much.
Due to the fact that borg allows you to mount archives as if they were file systems you can even pull out an individual message from the maildir if necessary. It's pretty neat.

Happy to take any feedback both on workability of this guide as well as writing style, spelling and grammar.

andryyy:
Sorry, I was so busy, I almost forgot it.

**Thank you!** :)

omexlu:
I use borg to backup my whole servers so I don't need a third party option, but I find that It will be a good idea to include this as standard into mailcow with more configuration options over the mailcow gui :)

Maybe a feature for feature :)

PR #271, opened 2021-05-01
Update third_party-borgmatic.md

sschwetz:
set crypt volume to be backed up to allow for bare-metal restore

### DIFF
--- a/docs/third_party-borgmatic.md
+++ b/docs/third_party-borgmatic.md
@@ -29,6 +29,7 @@ services:
     restart: always
     dns: ${IPV4_NETWORK:-172.22.1}.254
     volumes:
+      - crypt-vol-1:/mnt/source/crypt:ro
       - vmail-vol-1:/mnt/source/vmail:ro
       - mysql-socket-vol-1:/var/run/mysqld/:z
       - ./data/conf/borgmatic/etc:/etc/borgmatic.d:Z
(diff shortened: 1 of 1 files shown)


andryyy:
Hmm, not sure if we should do that. The fun part about having the maildir encrypted is easy encrypted mail backups.

We may want to comment it and add a note about it, what do you think?

Also pinging @kaechele

kaechele:
I think this actually addresses a major oversight on my end.
Borg backups are (unless otherwise instructed) encrypted locally before sent to the far end. They definitely are if set up the way I laid out in the documentation.
So also backing up the crypt volume actually would make sense for a bare metal restore.

For the sake of not producing unpleasant surprises when you actually do need to do full disaster recovery I would basically do a combination here:
- add the crypt volume by default
- add a note that for security reasons the crypt volume could also be backed up to a separate destination than the maildir and that this can be achieved by omitting the crypt volume from the docker-compose override.

docs/third_party-borgmatic.md
```diff
@@ -18,6 +18,10 @@ This guide only covers the basics.
 
 ### Create or amend `docker-compose.override.yml`
 
+!!! NOTE
+This setup will backup your crypt volume. This is to allow for full disaster recovery where you are installing onto bare tin. Borgmatic will dedupe and encrypt all data on your local machine before sending it across to the remote system.
+If for security reasons you do prefer to back up your crypt volume via another method remove  crypt-vol-1:/mnt/source/crypt:ro from the volumes in the yaml below.
```
kaechele, line 23: Putting the crypt-vol-1:/mnt/source/crypt:ro into backticks like this: `crypt-vol-1:/mnt/source/crypt:ro` makes it a little more obvious.

kaechele (changes requested):

PR #272, opened 2021-05-01
Create third_party-autoarchiving

sschwetz:
Setup autoarchving with dovecot-archive

### DIFF
--- a/docs/third_party-autoarchiving
+++ b/docs/third_party-autoarchiving
@@ -1,0 +1,100 @@
+# Automatic Archiving with dovecot-archive
+
+## Introduction
+
+dovecot-archive is a python wrapper for doveadm that allows for archiving email.
+
+This document guides you through the process to enable automated archiving of email across all mailcow domains. We will utilise ofelia to setup a schedule to trigger the archival script
+
+## Create or amend `docker-compose.override.yml`
+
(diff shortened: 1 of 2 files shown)


andryyy:
Hi, thank you! I don't think we need the ofelia container. :)

sschwetz:
I used the ofelia route as I already had the container running for other jobs and I prefer to keep the docker host’s crontab for the host. 

I will update the document with how to run it with a docker exec as well from corn

andryyy:
We already have ofelia in the stack.

sschwetz:
If that is the case should we also update https://mailcow.github.io/mailcow-dockerized-docs/u_e-dovecot-expunge/ ?

andryyy:
Yes! Good catch. :)

andryyy:
Thanks! Are you sure it's a cron compatible format? Iirc it also reads seconds. A `@every 4h` would be better imo.

I never checked that. I will merge and change it, ok?

andryyy:
Can you check if the changes are fine for you?

sschwetz:
I have updated a typo in the scheduler string, but it does need to have seconds declared

sschwetz:
@andryyy can you advise if you are happy with this now?

PR #273, opened 2021-05-02
Update Job Scheduling

sschwetz:
* Remove creation of ofelia container from docker-compose.override.yaml
* Remove cronjob schedule in lieu of ofelia-mailcow
* Fix grammatical errors


andryyy:
I also don't want to have the crontab part removed, I don't see a reason to do so. Hm.

PR #274, opened 2021-05-02
Update `pip install` command in readme

Daniel15:
Just installing `mkdocs-material` isn't sufficient. I copied the `pip install` command from the GitHub Actions config.


MAGICCC:
Yep indeed makes sense

PR #275, opened 2021-05-02
Document maildir_very_dirty_syncs

Daniel15:
Created a documentation page about `maildir_very_dirty_syncs`. I named it "Dovecot Performance Optimizations" so other performance-related settings could be documented there too.

Screenshot:
image

References https://github.com/mailcow/mailcow-dockerized/pull/4028

### DIFF
--- a/docs/u_e-dovecot-perf.md
+++ b/docs/u_e-dovecot-perf.md
@@ -2,7 +2,7 @@
 
 Dovecot's [`maildir_very_dirty_syncs` option](https://wiki.dovecot.org/MailLocation/Maildir#Optimizations) is enabled by default. This option can significantly improve the performance of mailboxes that contain very large folders (over 100,000 emails).
 
-What this option does is it avoids rescanning the entire `cur` directory whenever loading an email. With this option disabled, Dovecot takes it safe and scans the **entire** `cur` directory (essentially running an `ls`) to check if that particular email was touched (renamed, etc), by looking for all files whose names contain the correct ID. This is very slow if the directory is large, even on filesystems optimized for such use cases (such as ext4 with `dir_index` enabled) on fast SSD drives.
+What this option does is it avoids rescanning the entire `cur` directory whenever loading an email. With this option disabled, Dovecot takes it safe and scans the **entire** `cur` directory (comparable with running an `ls`) to check if that particular email was touched (renamed, etc), by looking for all files whose names contain the correct ID. This is very slow if the directory is large, even on filesystems optimized for such use cases (such as ext4 with `dir_index` enabled) on fast SSD drives.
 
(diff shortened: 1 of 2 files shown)


MAGICCC:
Would it make sense to add a warning box on top that you shouldn't fiddle with the files then?

andryyy:
Guys, it is **m**ailcow!

It does not really run an "ls", that would take like forever.
This option is about simultaneous writes.

As long as it's not 100% clear, we will not add it to the docs. :/

Daniel15:
> Guys, it is mailcow!

Oops! I'll update the page.

> It does not really run an "ls", that would take like forever.

I can reword the article, but that's essentially what it does though, and that's exactly why it takes forever for large maildirs 😛 

Please see the thread where the patch was originally posted: https://dovecot.org/list/dovecot/2009-March/037934.html, which says:

> This patch adds a new maildir_very_dirty_syncs setting. If set to "yes", Dovecot assumes it's the only one changing the cur/ directory (so other MDAs can add mails to new/ without problems). This makes it possible to avoid rescanning the cur/ directory all the time looking for filenames.

The thing is that some flags for maildir files are stored in the file name. For example, you could have a file named like this:
```
1620145567.M181443P1338726.b161ebb1591a,S=158325,W=161850:2,
```

If the email is marked as "seen", an `S` flag will be added to the end of the filename:
```
1620145567.M181443P1338726.b161ebb1591a,S=158325,W=161850:2,S
```

If the email is replied to, an `R` flag will be added:
```
1620145567.M181443P1338726.b161ebb1591a,S=158325,W=161850:2,RS
```

If you mark the email as seen or replied using a tool that directly modifies the maildir (ie. NOT through IMAP), the file will be renamed. In order for Dovecot to find this email again, it needs to scan the directory to find where the file moved to. Dovecot's implementation of this syncing is essentially just a `readdir` loop, but note that even a `glob` would require the entire directory index to be searched, which is effectively what `ls` does anyways.

To improve performance, what we can do is instead assume that nothing other than Dovecot is touching the maildir (specifically, the `cur` directories). If this is the case, this issue goes away - The file will never be renamed without Dovecot knowing about the rename (since it's Dovecot itself that's performing the renames), so there's no reason for it to rescan the directory. This is what `maildir_very_dirty_syncs` is 😃 

> This option is about simultaneous writes.

It doesn't really have anything to do with simultaneous writes, it's actually about other apps writing to the maildir (even if not simultaneous).

patschi:
To be picky, I'd suggest changing
```
essentially running an `ls`
```
to somewhat like:
```
comparable with running an `ls`
```

Otherwise: LGTM and might be ready to be merged when https://github.com/mailcow/mailcow-dockerized/pull/4028 is done.

Daniel15:
@patschi  Thanks! I'll make that change :)

Daniel15:
Why was this closed? It's still relevant.

PR #277, opened 2021-05-04
Dovecot default realm

sschwetz:
Explains how to set a default domain on dovecot to use if realm not supplied


andryyy:
I prefer to not have this in the docs. It is unsupported on our side and can be confusing (especially in combination with SOGo).

PR #279, opened 2021-05-05
borgmatic: Add archive prefixes / SELinux fixes

kaechele:
The borgmatic default for the archive prefix is the to use the hostname.
With docker-compose the hostname of the container will change on every
down/up cycle. So we introduce a static prefix instead.

Signed-off-by: Felix Kaechele <felix@kaechele.ca>

### DIFF
--- a/docs/third_party-borgmatic.md
+++ b/docs/third_party-borgmatic.md
@@ -62,7 +62,11 @@ location:
         - user@rsync.net:mailcow
     remote_path: borg1
 
+storage:
+    archive_name_format: "mailcow-{now:%Y-%m-%dT%H:%M:%S.%f}"
+
 retention:
+    prefix: mailcow
     keep_hourly: 24
     keep_daily: 7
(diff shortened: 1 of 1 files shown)


kaechele:
I just found out that I do need to label the vmail volume as shared (`z` vs `Z` volume flag).
Question is if this will compete with the `Z` flag on the vmail volume of the dovecot container in the main `docker-compose.yml`.

@andryyy Can we possibly mark the vmail volume as shared (`z` instead of `Z`)? I don't know what the behaviour of docker is if one container labels the volume as `Z` and the other one labels it as `z`.
The `backup_and_restore.sh` also labels the vmail volume as `z` on runtime, so it essentially creates the same problem.

derekslenk:
I would suggesting also adding a prefix to consistency, IE:

```yaml
consistency:
   prefix: mailcow
```
I came here to create that same pull request but with consistency added

PR #283, opened 2021-05-17
Added a warning for providers blocking SMTP Ports

holzhannes:
Hetzner is blocking outgoing traffic to ports 25 and 465 by default.


holzhannes (commented):
Improvments

MAGICCC:
Yes nice, but can you add the image to the repo like we did for the other images in the docs?
I also thought about a general info section under prerequirements, because there a lot of companies who are doing this (Hetzner, GCP, DO, Linode)

holzhannes:
@MAGICCC Thank you for your feedback. Sorry, I didn't see you are including images in another way, but I really like your Idea of a more generic warning. Maybe just listing providers who are doing it. I will change my PR, and maybe @andryyy could add an extra section for it later.

PR #286, opened 2021-05-25
Remove infinite redirect

john-montgomery2003:
Was having issues with infinite redirects, commenting out 
RewriteRule ^/?(.*) https://%{HTTP_HOST}/$1 [R=301,L]
[line61] seemed to fix the issue, and continues to redirect the http to https

### DIFF
--- a/docs/firststeps-rp.md
+++ b/docs/firststeps-rp.md
@@ -58,7 +58,7 @@ Let's Encrypt will follow our rewrite, certificate requests in mailcow will work
   RewriteEngine on
 
   RewriteCond %{HTTPS} off
-  RewriteRule ^/?(.*) https://%{HTTP_HOST}/$1 [R=301,L]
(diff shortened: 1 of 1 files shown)


andryyy:
Hm, the config works fine on multiple installations, Are you sure it's a problem with the RP config we provide?

It might introduce other problems when we merge this.

john-montgomery2003:
Cant be sure its this specific config, could be because of my other configs - I'll see if it persists on a fresh install, but i'll close for now.

andryyy:
Can you report back? Really curious. :) It might be important.

john-montgomery2003:
my bad - seems like there was some other issue, which i cant seem to isolate at the moment, it doesn't seem to be the fault of that line, but somehow commenting it fixed the issue I was facing :/ apologies for the wasted time

PR #295, opened 2021-08-04
Add LICENSE

maxcodeshere:
Material for MkDocs is distributed under the MIT license. Therefore this repo should contain the license file as well.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,19 @@
+Copyright (c) 2016-2021 Martin Donath <martin.donath@squidfunk.com>
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to
+deal in the Software without restriction, including without limitation the
+rights to use, copy, modify, merge, publish, distribute, sublicense, and/or
+sell copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NON-INFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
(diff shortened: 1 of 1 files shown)


MAGICCC:
Ah yes cool.
I wonder if we can put the info in the README that parts of the code we use is licensed under MIT.

andryyy:
This would probably also cover the content of the docs. 

I don’t think that’s something a foreign person should decide with a PR.

mkuron:
We should probably just add the same license as in the main repository: https://github.com/mailcow/mailcow-dockerized/blob/master/LICENSE . In fact, that was the license of the docs before we moved them to a separate repository.

andryyy:
What do you think, Max?

Max is close to the project, I did not know him by his GitHub name, sorry again.

MAGICCC:
I talked to him via Telegram and when we add a LICENSE file then its basicly for the  whole repo. I would still put the info in the README.md

maxcodeshere:
As I said when we talked: no offence taken André :) The PR was not completely thought through on my end and the intentions were not at all clear.

To clear things up: we'd have to differentiate between the licenses for the used software and the contents of the docs.

- Material for MkDocs is distributed under the MIT license. The MIT license states in part: _The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software._ 

- MkDocs itself is distributed under the BSD 2-Clause "Simplified" License. It states in part: _Redistributions of source code must retain the above copyright
notice, this list of conditions and the following disclaimer.
Redistributions in binary form must reproduce the above copyright
notice, this list of conditions and the following disclaimer in
the documentation and/or other materials provided with the
distribution._

- as of today André has a personal copyright notice for the contents of the docs in the page footer.


Sadly, what we cannot do, is what @mkuron suggested and just declare GPL for everything, as that would go directly against the requirements of MIT and BSD.

I think a practical solution might be either a paragraph in the README.md that states the licences of the components, as @MAGICCC  suggested or a separate document, e.g. acknowledgements.md, that contains the license texts for mkdocs and material for mkdocs.

What do you think would be a feasible way of doing it? I'll update the PR accordingly, once a decision is made :)

mkuron:
Is there even anything in this repository that is subject to MkDocs‘ copyright? As far as I can tell, we only have stuff here we wrote ourselves (like Markdown files). MkDocs itself is not in this repository, it is only used to convert Markdown to HTML. The authors of MkDocs have no copyright to the generated HTML either.

"Material for MkDocs" is a slightly different matter. That‘s the theme we use for the generated HTML, and of course the designer has rights to the design. So we would need to acknowledge them in a footnote in the HTML. But the theme itself is not contained in our docs repository, so it does not matter for the purpose of the license we use for the repository.

maxcodeshere:
That is actually a very good argument against my initial response.

I not entirely sure if an acknowledgement in the generated files would be necessary, as they could be considered to be files that are not distributed themselves trough the creator's repo. But I'll have look into that tomorrow


Regarding the design as it appears when rendered, I think it's fine as is. There already is an acknowledgement to the project in the footer of the rendered page.

Another thing I'll look into tomorrow is wether the content should be GPL or CC. I'll come back with a list of pros and cons :)

PR #296, opened 2021-08-18
Update traefik reverse proxy instructions

JanDevDE:
Mailcow cannot use port 80 internally as it's already being used by traefik by default. Thus it should be recommended to use 8080 in this docker-compose file to avoid port collision. This also follows the recommended reverse proxy mailcow.conf above "...Make sure you change HTTP_BIND and HTTPS_BIND in...."

### DIFF
--- a/docs/firststeps-rp.md
+++ b/docs/firststeps-rp.md
@@ -199,8 +199,9 @@ services:
         # Specifies which kind of cert resolver we'll use, in this case le (Lets Encrypt).
         - traefik.http.routers.moo.tls.certresolver=le
         # Creates a service called "moo" for the container, and specifies which internal port of the container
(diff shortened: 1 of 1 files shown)


MAGICCC:
I dont use Traefik myself but we could rather merge this PR because it uses the HTTP_PORT variable
https://github.com/mailcow/mailcow-dockerized-docs/pull/285

andryyy:
Oh, indeed. Feedback is welcome. :)

JanDevDE:
@MAGICCC I can confirm that HTTP_PORT mentioned in #285 works.

MAGICCC:
OK cool thanks.
I merged the other one and close this.

PR #299, opened 2021-09-08
Update u_e-mailcow_ui-tagging.md

JXGA:
Include the term 'plus addressing' and the RFC 5233 term 'sub-addressing' to make the information easier to find when searching.


andryyy:
Thank you. :)

PR #300, opened 2021-09-13
add /opt/mailcow-dockerized to borgmatic example

qupfer:
Hi,
to backup modifications on mailcow.conf (and more, like nginx redirects, roundcube...) /opt/mailcow-dockerized should be part of the backup too.


MAGICCC:
Hm, then you basicly backup the whole mailcow repo too. Does it also include the .git folder?

qupfer:
> Hm, then you basicly backup the whole mailcow repo too. Does it also include the .git folder?

Yes, is that bad? I think the size is not much in comparison to a typical mail dir.
Otherwise there are possibilities to exclude the .git folder. And will the update script work without .git?

milkmaker:

PR #302, opened 2021-09-30
Update Roundcube doc to fix spellcheck config

apharp:
Changing from "pspell" to "aspell"


FingerlessGlov3s:
Can confirm this fixes it.

PR #306, opened 2021-10-14
Guidance on setting up IPv6

joseluisgp:
I've run into issues when setting up mailcow on several providers when wanting to use an specific IPv6 address for all mail services, as some don't have IPv6 enabled out of the box or don't have all IPs in the assigned range on the list of addresses in the system config. I'm assuming the reader uses a Debian-bases system with netplan.


milkmaker:

PR #336, opened 2021-12-22
third_party-borgmatic: Improve volume configuration

ramcq:
Some updates to the recommended borgmatic configuration.

Try to more closely match the backup_and_restore.sh script by including postfix, rspamd and redis volumes. Ignore the paths that contain sockets/FIFOs as borgmatic uses these for its mysql backup, which means borg "stuck" trying to read these.

Move the generated borg security data and cache to volumes so that they persists without writing in the mailcow/data folder or the overlay.

### DIFF
--- a/docs/third_party-borgmatic.md
+++ b/docs/third_party-borgmatic.md
@@ -23,17 +23,23 @@ configuration:
 
 ```yaml
 version: '2.1'
+
 services:
   borgmatic-mailcow:
     image: b3vis/borgmatic
+    hostname: mailcow
     restart: always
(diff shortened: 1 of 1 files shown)


ramcq:
Ah, also adding a hostname inside the container makes the borg backup IDs in your repo stable across container rebuilds. Otherwise your backups have a name like "242dcb16fbce-..." and next time you rebuild the container, it doesn't recognise the old ones / rotate them correctly.

MAGICCC:
thanks!

PR #341, opened 2022-01-04
Revert "Update third_party-roundcube.md"

sandrzejewskipl:
Reverts mailcow/mailcow-dockerized-docs#317

Why? Because if u add this piece of code at the bottom of the file, it's throwing internal error. After removing <?php, it's working.

### DIFF
--- a/docs/third_party-roundcube.md
+++ b/docs/third_party-roundcube.md
@@ -67,8 +67,6 @@ Initialize the database and leave the installer.
 
 Open `data/web/rc/plugins/managesieve/config.inc.php` and change the following parameters (or add them at the bottom of that file):
 ```
-//Make sure you have <?php at the beginning of the file or add just like below
-<?php 
 $config['managesieve_port'] = 4190;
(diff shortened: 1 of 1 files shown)


MAGICCC:
Well see the commented code
```
//Make sure you have <?php at the beginning of the file or add just like below
```

I dont know how the org file looks like since I dont use Roundcube

sandrzejewskipl:
> Well see the commented code
> 
> ```
> //Make sure you have <?php at the beginning of the file or add just like below
> ```
> 
> I dont know how the org file looks like since I dont use Roundcube

https://github.com/roundcube/roundcubemail/blob/master/plugins/managesieve/config.inc.php.dist

If u are only changing values, it doesn't matter, because u are not doing copy paste. When u are copy pasting this piece of code, at the end of the file, another <?php is a problem.

sandrzejewskipl:
like there is no chance to not be there <?php in this config, so i dunno how this PR was supposed to help

MAGICCC:
Indeed... thanks

PR #343, opened 2022-01-05
Roundcube remote resources fix and tidy

FingerlessGlov3s:
Due to how mailcow is configured security wise, some people may get this issue https://github.com/roundcube/roundcubemail/issues/8170 depending on their browser. When you go to press `Allow` to show remote content within an email, it'll redirect you to the root of the website, in this case mailcow login. This fix should be included in the Roundcube 1.6 release, until we hit this release we just need to apply the fix manually for each install/upgrade.

### DIFF
--- a/docs/third_party-roundcube.md
+++ b/docs/third_party-roundcube.md
@@ -5,11 +5,15 @@ Download Roundcube 1.5.x to the web htdocs directory and extract it (here `rc/`)
 # Check for a newer release!
 cd data/web
 wget -O - https://github.com/roundcube/roundcubemail/releases/download/1.5.2/roundcubemail-1.5.2-complete.tar.gz | tar xfvz -
+
 # Change folder name
 mv roundcubemail-1.5.2 rc
 
 # Change permissions
 chown -R root: rc/
+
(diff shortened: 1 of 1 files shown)


MAGICCC:
Or we tell them to download directly the file from the commit
https://raw.githubusercontent.com/roundcube/roundcubemail/28b635175090666f8d9f68a55ae11091ec00b0b6/program/include/rcmail.php

Theres another if else in the else

FingerlessGlov3s:
I've used roundcube with this fix since September, so I know it works fine.

I just squashed the "patch" https://github.com/roundcube/roundcubemail/issues/8170#issuecomment-922436182 in to one line, to make the SED easier.

We can pull down the files from that commit indeed but the SED solution is known to work and simpler and hopefully won't need to do it for long. We only need to caterer for it working in mailcow's environment.

Otherwise I need to compare the versions of those files shipped in 1.5.2 with the ones for that commit to see if there are any other changes, since we do not want to reintroduce bugs that have been fixed in recent releases after 13 November 2021.

MAGICCC:
It's ok :)
When RC released 1.6. and they added the fix we need to remove this line again

PR #361, opened 2022-02-04
Disable prefix in Borgmatic retention

BBaoVanC:
By default, Borgmatic uses `{hostname}-` as the prefix, and creates backups with that prefix. In Docker, the hostname of the container will change every time the container is recreated, so under default settings, Borgmatic will noty prune archives that were created in a different container. This change disables the prefix so Borgmatic will prune any archive according to the retention policy.


DerLinkman:
Thanks :)

Issue #367, opened 2022-02-16
nextcloud cron task


BigMichi1:
as Nextcloud needs to perform some cron related actions the documentation https://mailcow.github.io/mailcow-dockerized-docs/third_party/third_party-nextcloud/ should be extended to include a part on how to configure/enable the cron task.

see https://docs.nextcloud.com/server/latest/admin_manual/configuration_server/background_jobs_configuration.html

I achieved it by adding this part to my `docker-compose.override.yml` and running `docker-compose up -d && docker-compose restart ofelia`

```yaml
version: '2.1'
services:
  php-fpm-mailcow:
    labels:
      ofelia.enabled: "true"
      ofelia.job-exec.nextcloud-cron.schedule: "@every 5m"
      ofelia.job-exec.nextcloud-cron.command: "su www-data -s /bin/bash -c \"/usr/local/bin/php -f /web/nextcloud/cron.php\""
```

MAGICCC:
If you want you can create a PR for that

PR #370, opened 2022-02-18
Nextcloud background job execution

BigMichi1:
adding a new section about how to add background job execution with `ofelia` for Nextcloud

closes: #367


MAGICCC:
thanks

PR #382, opened 2022-03-23
Fix link to point to new docs location

RafaelKr:


MAGICCC:
Even that we added redirects of all sites to the new layout, I'll merge it

RafaelKr:
Then you missed a redirect for https://mailcow.github.io/mailcow-dockerized-docs/b_n_r_backup/
I get a 404 for that page.

MAGICCC:
Ah yes right...
Fixed in https://github.com/mailcow/mailcow-dockerized-docs/commit/3f82c5437d4940ca3e3041b0c9f523884e30342f

PR #392, opened 2022-04-08
Increase ReceiveTimeout value.

accolon:
Download speed for free SecuriteInfo databases is limited to 300 kB/s, the default timeout of 20 sec is too low for some of the larger databases (like javascript.ndb with 15 MB) so downloads will fail.

Fixed a few language issues as well.


DerLinkman (approved):
Looks good.

Thanks for that!

PR #395, opened 2022-04-15
Add cert info to third_party nextcloud doc

dubwoc:
Add instructions around configuring ADDITIONAL_SAN for the Nextcloud domain with a reference to more detailed documentation.


DerLinkman (approved):
Looks good to me now.

I´ve added the german version of that as well to be consistent with the rest.

PR #407, opened 2022-05-19
Added note to DNS setup configuration

piuswalter:
Hello all,

I have added a small hint to the DNS setup documentation, because the `mail` DNS record which binds the subdomain to the given ip address must only be set for the domain on which mailcow is running and that is used to access the web interface.

If this DNS recotd is created for another domain for which email addresses exist in mailcow, it will only result in a certificate warning by calling the `mail` subdomain.


DerLinkman (approved):
Looks good :)

Thanks for the addition

PR #409, opened 2022-05-24
Added section to check rsyslog listens on port 514 udp

myOmikron:
As rsyslog doesn't listen on 514 UDP with the default configuration file shipped with Debian 10 and 11, I added a section to enable syslog reception via UDP.

### DIFF
--- a/docs/post_installation/firststeps-logging.de.md
+++ b/docs/post_installation/firststeps-logging.de.md
@@ -68,16 +68,27 @@ services:
       options:
         syslog-address: "udp://127.0.0.1:514"
         syslog-facility: "local3"
+```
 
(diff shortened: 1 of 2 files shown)


myOmikron:
I guess it would also be a good idea to add a section regarding logrotation. Feel free to intervene

DerLinkman:
> I guess it would also be a good idea to add a section regarding logrotation. Feel free to intervene
> 
> 

Yes sounds good. But please do that in a separate Pull Request.

myOmikron:
> Looks good so far. Does this made rsyslog listen on that desired port?

Yes, rsyslog will start listening on `0.0.0.0:514` and `[::]:514`.

PR #410, opened 2022-05-24
Added section to describe setup of logrotation

myOmikron:
Follow-up to #409


DerLinkman (approved):
Looks good!

Thanks for that.

PR #447, opened 2022-07-28
Switch docs' instructions to Docker Compose v2

raph-topo:
Search-replace `docker-compose` for `docker compose`.
Manual proofreading of all changes.


DerLinkman:
Did you´ve updated mailcow to a version newer than 2022-06?

The reason I ask is this: We did this in the update.sh script etc. for a reason and left it as it is in the current docs. Say that we docker-compose so standalone after install and not the version from the repo or the plugin.

raph-topo:
Hello, 
I was about to answer that mixing syntax of Compose v1 & v2 would create confusions, but you resolved it in a69cfb806f1ab02ae3bd6cd772b61d663340661f 👍 👍

PR #448, opened 2022-07-28
Update Roundcube docs page going from 1.5 to 1.6

FingerlessGlov3s:
This is to update the docs page to install or upgrade to Roundcube 1.6.0 along with the needed 'sed' commands to make any necessary config changes going from 1.5 to 1.6

### DIFF
--- a/docs/third_party/roundcube/third_party-roundcube.de.md
+++ b/docs/third_party/roundcube/third_party-roundcube.de.md
@@ -1,19 +1,16 @@
 ## Installation von Roundcube
 
-Laden Sie Roundcube 1.5.x in das Web htdocs Verzeichnis herunter und entpacken Sie es (hier `rc/`):
+Laden Sie Roundcube 1.6.x in das Web htdocs Verzeichnis herunter und entpacken Sie es (hier `rc/`):
 ```
 # Prüfen Sie, ob eine neuere Version vorliegt!
 cd daten/web
-wget -O - https://github.com/roundcube/roundcubemail/releases/download/1.5.3/roundcubemail-1.5.3-complete.tar.gz | tar xfvz -
+wget -O - https://github.com/roundcube/roundcubemail/releases/download/1.6.0/roundcubemail-1.6.0-complete.tar.gz | tar xfvz -
 
 # Ändern Sie den Ordnernamen
-mv roundcubemail-1.5.3 rc
+mv roundcubemail-1.6.0 rc
 
 # Berechtigungen ändern
 chown -R root: rc/
-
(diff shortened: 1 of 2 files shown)


MAGICCC:
Could you also edit the german variant of the file?

FingerlessGlov3s:
Of course I can, can you or someone translate the comment bit? As I can't speak German and I fear google translate will do it wrong.

MAGICCC:
Ah right

```diff
- # Type 'N' to "Do you want me to fix your local configuration" if prompted
+# Geben Sie 'N' ein, wenn folgender Dialog erscheint: "Do you want me to fix your local configuration".
````

```diff
- # If your going from 1.5 to 1.6 please run the config file changes below
+# Falls Sie von Version 1.5 auf 1.6 updaten, dann führen Sie folgende Befehle aus, um die Konfigurationsdatei anzupassen:`
```

accolon:
When doing an upgrade, the `installto.sh` script shows `NOTICE: Update dependencies by running 'php composer.phar update --no-dev'`.

I always do this in the `/web/rc` folder afterwards which usually updates a few packages, especially after upgrading from 1.5 to 1.6.

Should this be added to the instructions?

FingerlessGlov3s:
@accolon, I didn't get that notice or at least remember it, have you installed any extra plugins?

accolon:
I'm only using the plugins mentioned in the docs, e.g. `carddav` or `password`.

Composer also updates dependencies which are part of Roundcube, like guzzle or bacon-qr-code, since the `composer.json` which is part of the "complete" RC tar.gz allows higher versions for those.

I didn't write down the changes it did this time, I will keep an eye on it. I just always updated the dependencies and was wondering if this should be added to the instructions.

FingerlessGlov3s:
We could do the compose updates but then the dependencies wouldn't match the released complete version, which I'm not 100% sure we an want to do. I would think the ones released with Roundcube will be known good from the developers.

The official upgrade instructions only mention doing that if doing the upgrade manually. 🤔
https://github.com/roundcube/roundcubemail/blob/master/UPGRADING

accolon:
Fine by me. You're right with regard to the upgrade instructions.

On the other hand, the `update.sh` script which is called by `installto.sh` seems to rebuild `composer.json` each time and it always prints out the notice about updating dependencies...

But since it has worked without this additional step for quite some time, it is probably best to keep it this way.

accolon:
Oh, one additional thing:

They also renamed `smtp_server` to `smtp_host` in 1.6:
https://github.com/roundcube/roundcubemail/commit/893216cb297268d222ae49099e6654a304b72e3f

FingerlessGlov3s:
Hehe the SED is correct but not the example config on first install, good spot.
I'll fix that

MAGICCC:
thanks :)

PR #466, opened 2022-09-11
Email decryption to use correct compression level

vangeance666:
As of this Dovecot [commit](https://github.com/dovecot/core/commit/9b420dde6e3df30fbadd55e4e8524c809437ee03), it doesn't permit compression level of 0.


DerLinkman (approved):
Merci :)

PR #477, opened 2022-10-02
updated docs for install to 1.6.0 - added composer update stanza

bastischubert:
After updating to Roundcube 1.6.0 i was not able to Send mails anymore, due to a missing PEAR class 
``` 
PHP Fatal error:  Uncaught Error: Call to undefined method Net_SMTP::starttls() in /web/rc/program/lib/Roundcube/rcube_smtp.php 
```
After looking at the output from the installto script i noticed that i had to update the dependencies.

This commit shares that information with other users ;)

Signed-off-by: Sebastian Schubert <basti@2-die-4.tk>


DerLinkman:
Merci :)

PR #484, opened 2022-10-19
Improve customized Dockerfile

Skydiver84de:
I just tried to enhance the customized Dockerfile with my personal experience after just doing this for a customized watchdog container. Personalized $tag and container activation in docker-compose.override.yml was added. Without that the customized container was not used after docker-compose up -d (at least for me).


DerLinkman:
Thank you!

PR #489, opened 2022-11-04
Update borgmatic

NicoDreamzZ:


NicoDreamzZ:
ssh repositorys should be configured like this: ssh://user@backupserver/./sourcehostname.borg 
see: https://torsion.org/borgmatic/docs/reference/configuration/

I added the port so nobody needs to search where to configure it.

DerLinkman (approved):
Thank you!

PR #491, opened 2022-11-04
Add explanation for testing with placeholder values replaced

bprfh:
I think it would be a good idea to add an explanation on how the variables in the documentation are replaced and how to preview changes with "real" values filled in.

I closed the first pull request because the example url was not working, sorry about that.


DerLinkman (approved):
Thanks for that!

PR #492, opened 2022-11-04
checkmk cache

FingerlessGlov3s:
This PR is to add information about using the cache ability in checkmk, this is so the check isn't ran every time the agent is checked. The default check interval is 1 minute for checkmk.

To reduce server load when the check happens, it makes sense to cache the result and only run the actual check every hour or so.


FingerlessGlov3s:
Forgot to add, that the German may need amending since I used google translate.

DerLinkman (approved):
Looks good ;) Thanks a lot!

PR #493, opened 2022-11-08
Update firststeps-dmarc_reporting.en.md

dragoangel:


docs/post_installation/firststeps-dmarc_reporting.en.md
```diff
@@ -45,8 +45,8 @@ services:
       - MASTER=${MASTER:-y}
     labels:
       ofelia.enabled: "true"
-      ofelia.job-exec.rspamd_dmarc_reporting.schedule: "@every 24h"
-      ofelia.job-exec.rspamd_dmarc_reporting.command: "/bin/bash -c \"[[ $${MASTER} == y ]] && /usr/bin/rspamadm dmarc_report > /var/lib/rspamd/dmarc_reports_last_log 2>&1 || exit 0\""
+      ofelia.job-exec.rspamd_dmarc_reporting_yesturday.schedule: "@every 24h"
```
DerLinkman, line 48: I think here and in the next row is a typo: yesterday instead of yesturday

docs/post_installation/firststeps-dmarc_reporting.en.md
```diff
@@ -106,13 +106,33 @@ docker compose exec redis-mailcow redis-cli ZRANGE "dmarc_rpt;microsoft.com;mail
...
+```
+version: '2.1'
+
+services:
+  rspamd-mailcow:
+    environment:
+      - MASTER=${MASTER:-y}
+    labels:
+      ofelia.enabled: "true"
+      ofelia.job-exec.rspamd_dmarc_reporting_yesturday.schedule: "@midnight"
```
DerLinkman, line 128: Here as well: yesterday instead of yesturday.

DerLinkman (changes requested):
View comments :)

dragoangel:
Thank you :)

What about the text itself? Is it clear enough?

DerLinkman:
Yeah the text is fine.

Thanks

DerLinkman (approved):
Looks good :)

wmtech-1:
If I would be very picky, I wouldn't run the "yesterday" report (when running the reports more frequently) exactly at midnight. Because then it could happen that one or two messages processed within the last seconds of yesterday don't get dmarc reports. But that should not be too problematic. ;-)

To be on the safe side, you could change `"@midnight"` to `0 10 0 * * *`. 

But again, this is very picky of me.

PR #497, opened 2022-11-19
Manual config urls should use client vars

bprfh:
In the client-manual config page the example urls use placeholder values like username@mail.example instead of the values specific to the installation.
I often have to setup manual clients and have to manually replace these values instead of just copying the manual links.
This pr fixes that and uses the username and host addres instead of the example values if available.


DerLinkman (approved):
Thank you!

PR #507, opened 2022-12-31
Add a reminder for SNAT_TO_SOURCE variable after migration

ooguz:
I have moved our mailcow server from one server to another by following this tutorial, but [mailcow/mailcow-dockerized#4627](https://github.com/mailcow/mailcow-dockerized/issues/4627) has occured, and the problem was the misconfiguration for this variable, I've spent around half an hour on investigation. A reminder has added for this situation in order to prevent some tears :)

After the migration, this variable was pointing the old IPv4 address of the server, thus SOGo could not make IMAP connections; after changing it to the new one the issue was fixed.

### DIFF
--- a/docs/i_u_m/i_u_m_migration.de.md
+++ b/docs/i_u_m/i_u_m_migration.de.md
@@ -44,4 +44,4 @@ docker compose pull
 docker compose up -d
 ```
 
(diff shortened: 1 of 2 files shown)


DerLinkman (approved):
Thank you!

PR #509, opened 2023-01-02
Fix docker compose volume warnings in borgmatic config

NicoDreamzZ:
This fixes the warnings that appear when executing docker compose up -d

```
WARN[0000] mount of type `volume` should not define `bind` option
WARN[0000] mount of type `volume` should not define `bind` option
WARN[0000] mount of type `volume` should not define `bind` option
WARN[0000] mount of type `volume` should not define `bind` option
```

It seems docker compose don't like the :Z option on volumes

This should be similar to https://github.com/mailcow/mailcow-dockerized/pull/4577


DerLinkman (approved):
Thank you!

PR #528, opened 2023-02-02
Added new gitea database

fabreg:
Excluded gitea database from mailcow


MAGICCC:
Thanks, can you also do the same for the Gogs docs? I will try to provide a german translation later.

fabreg:
Sure.
https://github.com/mailcow/mailcow-dockerized-docs/pull/530

p.s. not tested yet cause I'm not using Gogs (but it's supposed to work)

MAGICCC (approved):

PR #534, opened 2023-02-15
install: fix package name on debian/ubuntu

sebix:


DerLinkman:
Could you at least provide a short description why this should be implemented?

I can't see a reason yet as the compose plugin is installed with that exact syntax on debian/ubuntu.

sebix:
You are right, the package name `docker-compose-plugin` is intentional.

PR #554, opened 2023-03-27
Improve paragraph "check the certificate served by Postfix"

raph-topo:
Label STARTTLS ports as STARTTLS.
Add checks for TLS ports.
Improve command.


raph-topo:
Look like Github forgot this had already been merged…

PR #560, opened 2023-04-12
Update client-thunderbird.en.md

Marethyu9999:
adding a optional step to sync all folders, instead of only the subscribed ones. This should be especially useful if the "Set handling for tagged mail:" is set to "in subfolder"


Marethyu9999:
should I do another pull request for the german entry or should I just add it here?

MAGICCC (approved):

PR #571, opened 2023-05-10
Update roundcube install instructions

mstilkerich:
Hello,

 I am in the process of migrating my mail server to mailcow, and as part of it I installed roundcube because it is my preferred webmailer.

Originally I just intended to update the mailcow instructions regarding the CardDAV plugin (of which I am the author), because the current version 5 supports some features like rediscovery of added or removed addressbooks on the server that are of interest in combination with SOGo, but I thought other parts of the setup could be improved as well, so the changes are a bit bigger.

Main changes:
- Use separate database for roundcube
- Download mime.types once and not on every container start
- Use mailcow API for password change from roundcube instead of direct database access
- Use non-TLS connection instead of TLS with disabled certificate check
- Upgrade CardDAV instructions to use RCMCardDAV v5 which also includes rediscovery of new/removed server-side addressbooks
- Instructions for forwarding client IP to dovecot (main benefit: netfilter container acts on excessive failed logins)
- Composer is already part of the php-fpm container, so we can use it to install plugins and also no need to download it for updating roundcube.

I did not try the admin login part into roundcube because it is of no interest to me, so I just kept the existing text. I also did not try the upgrade part but it looks fine to me from what I am used to when upgrading roundcube.

I can also update the German instructions, but I would first like to gather feedback on the English version and update the German one when there is agreement on the English variant.

### DIFF
--- a/docs/third_party/roundcube/third_party-roundcube.de.md
+++ b/docs/third_party/roundcube/third_party-roundcube.de.md
@@ -15,7 +15,7 @@ Das roundcube-plugin-installer composer Plugin hat eine [Design-Schwäche](https
 die dazu führen kann, dass composer bei Operationen fehlschlägt, im Rahmen derer Pakete aktualisiert oder deinstalliert
 werden.
 
-Die Fehlermeldung in diesem Falle besagt, dass eine `require`-Anweisung in `autoload_real/php` fehlgeschlagen ist, weil
+Die Fehlermeldung in diesem Falle besagt, dass eine `require`-Anweisung in `autoload_real.php` fehlgeschlagen ist, weil
 eine Datei nicht gefunden werden konnte. Beispiel:
 
 ```
@@ -235,7 +235,7 @@ services:
 Das Ändern des mailcow Passworts aus der Roundcube-Benutzeroberfläche wird durch das password-Plugin ermöglicht. Wir
 konfigurieren dieses zur Verwendung der mailcow-API zur Passwort-Aktualisierung, was es zunächst erfordert, die API zu
 aktivieren und den API-Schlüssel zu ermitteln (Lese-/Schreib-Zugriff notwendig). Die API kann in der
-mailcow-Administrationsoberfläche aktiviert werden, so Sie auch den API-Schlüssel finden.
+mailcow-Administrationsoberfläche aktiviert werden, wo Sie auch den API-Schlüssel finden.
 
 Öffnen Sie `data/web/rc/config/config.inc.php` und aktivieren Sie das Passwort-Plugin, indem Sie es dem
 `$config['plugins']`-Array hinzufügen, zum Beispiel:
@@ -251,7 +251,7 @@ $config['plugins'] = array(
 );
 ```
 
-Konfigurieren Sie das password-Plugin (stellen Sie sicher, __**API_KEY**__ auf Ihren mailcow Lese-/Schreib-API-Schlüssel
+Konfigurieren Sie das password-Plugin (stellen Sie sicher, __\*\*API_KEY\*\*__ auf Ihren mailcow Lese-/Schreib-API-Schlüssel
 anzupassen):
 
 ```bash
@@ -598,7 +598,7 @@ EOCONFIG
 ### Umstellung des RCMCardDAV-Plugins auf die Installation mittels composer
 
 Dieser Schritt ist optional, aber er gleicht Ihre Installation an die aktuelle Fassung der Anweisungen an und ermöglicht
-die Aktualisierung von RCMCardDAV mittels composer. Dies wird einfach dadurch erreicht, indem das carddav-Plugin aus dem
+die Aktualisierung von RCMCardDAV mittels composer. Dies wird einfach dadurch erreicht, dass das carddav-Plugin aus dem
 Installationsverzeichnis gelöscht und entsprechend der [Anweisungen oben](#CardDAV-Adressbücher-in-Roundcube-einbinden)
 installiert wird, einschließlich der Erstellung einer neuen RCMCardDAV v5-Konfiguration. Falls Sie das RCMCardDAV
 angepasst haben, sollten Sie dieses sichern, bevor Sie das Plugin löschen, und Ihre Anpassungen später in die neue
--- a/docs/third_party/roundcube/third_party-roundcube.de.md
+++ b/docs/third_party/roundcube/third_party-roundcube.de.md
@@ -1,206 +1,375 @@
 ## Installation von Roundcube
 
-Laden Sie Roundcube 1.6.x in das Web htdocs Verzeichnis herunter und entpacken Sie es (hier `rc/`):
-```bash
-# Prüfen Sie, ob eine neuere Version vorliegt!
-cd data/web
-wget -O - https://github.com/roundcube/roundcubemail/releases/download/1.6.1/roundcubemail-1.6.1-complete.tar.gz | tar xfvz -
+Sofern nicht abweichend angegeben wird für alle aufgeführten Kommandos angenommen, dass diese im mailcow
+Installationsverzeichnis ausgeführt werden, d. h. dem Verzeichnis, welches `mailcow.conf` usw. enthält. Bitte führen Sie
+die Kommandos nicht blind aus, sondern verstehen Sie was diese bewirken. Keines der Kommandos sollte einen Fehler
+ausgeben; sollten Sie dennoch auf einen Fehler stoßen, beheben Sie diesen sofern notwendig bevor Sie mit den
+nachfolgenden Kommandos fortfahren.
 
-# Ändern Sie den Ordnernamen
-mv roundcubemail-1.6.1 rc
+### Hinweise zur Verwendung von composer
 
-# Berechtigungen ändern
-chown -R root: rc/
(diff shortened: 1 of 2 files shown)


MAGICCC:
It looks fine for me thanks! Since I don't use roundcube, I will tag @FingerlessGlov3s, since he was the one who kept updating the docs for Thunderbird

FingerlessGlov3s:
Thanks for the ping @MAGICCC

Some good work here @mstilkerich

I've reviewed the updated docs, it's quite the overhaul. I haven't currently got a mailcow test instance to follow the instructions start to finish. I do like the idea of moving round cube to it's own database.
If we're to change how the setup is done, I feel we need an upgrade path of instructions for users who currently do it the current way, to try keep all installations the same configuration, will stop us needing to cater for multiple upgrades methods in the future should a breaking change happen, we may need to write 2 upgrade instructions. We maybe able to simply copy the tables over, remove table prefix, remove the current install directory and then do the installation missing out the running of `/installer/` step somewhat.
Other method is just to ask people to kindly do a complete reinstall, but I know some people won't want to loose any round cube data saved in the database like contacts or settings.

We need a nginx rule to disable accessing of `/web/rc/temp/` and `/web/rc/logs/` to stop people being able to publicly download any files in those directories. I think this is easily done with a simple `.conf` file

Changing it to only download the mime.types once, we should be getting users to download a new version every upgrade, so we are keeping it somewhat up to date. So any new types get added over time.

The `Disable and remove installer` section needs little bit of rewording just to say it needs changing to `false`. Something like

> Delete the directory `data/web/rc/installer` after a successful installation, and set the `enable_installer` option to `false` in `data/web/rc/config/config.inc.php`!

I don't use the `Let admins log into Roundcube without password` feature either, so can't comment on it.

The upgrade instructions will still work, as they are currently for both installation methods current and this one.

mstilkerich:
Thanks for the great feedback, especially regarding exposing the logs/temp folders.

I added an nginx config snippet that only exposes the `public_html` subdirectory of roundcube, so it should not be possible to access anything in `rc/` except what is in `rc/public_html`. I would appreciate feedback regarding this, since I have only used Apache so far and am completely new to nginx.

I also made some further adjustment, that do not need to be reviewed yet as I will probably change some more. I'll also think about a migration description.

I'm on a trip the next two weeks so it may take a while for me to follow up, but I will follow up.

FingerlessGlov3s:
I think it would also be worth cleaning up the old tables in mailcow database after we've dumped and imported into the new database.

I shall test this installation and migration guide once we think we are confident it's ready :-)

mstilkerich:
Thanks for the feedback. I have incorporated your suggestions.

mstilkerich:
Hi, did you find some time to test the steps?

FingerlessGlov3s:
I've managed to find some time 😀

I've tested the whole guide, there's a things that I've had to do to get things working or things I would change.

# Testing Fresh Install of Roundcube

In the roundcube configuration.
Ideally we shouldn't be using CBC ciphers. Please change the cipher_method from `AES-256-CBC` to `chacha20-poly1305` which is also an AEAD cipher. The default is even worse, I should of changed this in the past! 😑

When configuring the password plugin, it errors when trying to change password because it looks to be redirecting and then the hostname doesn't match the HTTPS certificate.
To fix this I set my mailcow host to the external domain name, which got it working for me. Fix below
```
cat <<EOCONFIG >data/web/rc/plugins/password/config.inc.php
<?php
\$config['password_driver'] = 'mailcow';
\$config['password_confirm_current'] = true;
\$config['password_mailcow_api_host'] = 'https://${MAILCOW_HOSTNAME}';
\$config['password_mailcow_api_token'] = '**API_KEY**';
EOCONFIG
```

When I tried to install the CardDAV intergration, I had to add `-W` to get the dependencies to install
```
docker exec -it -w /web/rc $(docker ps -f name=php-fpm-mailcow -q) composer require --update-no-dev -W -o "roundcube/carddav:~5"
```
Then when I tried to access roundcube to test there's missing tables, so it doesn't work. \
```
<11b04ef2> DB Error: [1146] Table 'roundcubemail.carddav_accounts' doesn't exist (SQL Query: SELECT * FROM carddav_accounts WHERE ((`user_id` = '2'))) in /web/rc/program/lib/Roundcube/rcube_db.php on line 567 (GET /rc/)
```
A quick google or two didn't reveal any quick fixes so I skipped the testing of CardDav.

I didn't test `Let admins log into Roundcube without password`, since I'm not really interested in this bit myself.
If I need to login to someone's mailbox I just add a app password and login.

Ended with a fully working installation of Roundcube

# Testing inplace upgrade from old method to new method
When doing the migration of tables, we can remove the prefix by doing a simple sed replace. Then remove that bit at the bottom about messing with that setting.
```
docker exec $(docker ps -f name=mysql-mailcow -q) /bin/bash -c "mysqldump -uroot -p${DBROOT} mailcow $RCTABLES | sed 's/mailcow_rc1//' | mysql -uroot -p${DBROOT} roundcubemail"
```

Then use sed again to remove the no longer required `db_prefix`, `log_dir` and `temp_dir`, so the same folders are used.
```
sed -i "/\$config\['db_prefix'\].*$/d" data/web/rc/config/config.inc.php
sed -i "/\$config\['log_dir'\].*$/d" data/web/rc/config/config.inc.php
sed -i "/\$config\['temp_dir'\].*$/d" data/web/rc/config/config.inc.php
```

I also then did the following to bring over some of the newer settings and the web folder protection.
```
cat <<EOCONFIG >>data/web/rc/config/config.inc.php
\$config['cipher_method'] = 'chacha20-poly1305';
EOCONFIG

cat <<EOCONFIG >data/conf/nginx/site.roundcube.custom
location /rc/ {
  alias /web/rc/public_html/;
}
EOCONFIG
```

Once I confirmed I had migrated I dropped the old tables, as the guide says at the bottom. I feel this should be in the migration section, as if it works we should clean up.
```
docker exec -it $(docker ps -f name=mysql-mailcow -q) mysql -uroot -p${DBROOT} -sN mailcow -e "SET SESSION foreign_key_checks = 0; DROP TABLE IF EXISTS $(echo $RCTABLES | sed -e 's/ \+/,/g');"
```

When I did migrate my roundcube I still used the TLS connections to dovecot, postfix and sieve. I do get that it's all local traffic, but saves me disabling security bits.

mstilkerich:
Thanks a lot for trying this and your feedback which contains a couple of points I missed. I have a few questions regarding the issues you encountered:

1) Redirection of http -> https: Are you maybe using a modified nginx config as described [here](https://docs.mailcow.email/manual-guides/u_e-80_to_443/)? It does not happen in my case (I run mailcow behind a reverse proxy, I guess many people do). I would prefer to have containers directly talk to each other without passing through the NAT tables / reverse proxy on the host. If you can confirm it only happens with a modified nginx config, how about adding a note that when having the HTTPS redirection configured, the public hostname should be used?

2) The `-W` option to composer when installing plugins might be a good idea anyway, though I wonder why the issue did not occur for me since I guess we both had the roundcube 1.6.1 complete package with the shipped composer dependencies as a starting point. I will add it.

3) The carddav plugin creates / updates the database tables when logging into roundcube. Did you maybe already have a roundcube session running before installing the carddav plugin and then did not log out of roundcube when testing it? This is my only explanation. Anyway, I will add a note that the login to roundcube is needed, but maybe you can confirm this actually resolves the issue on your side as well.

4) I did intentionally not change the cipher method in the migration instructions since it will break decryption of any stored emcrypted data migrated from the old database, for example all the CardDAV passwords. It might affect data stored at the client side in cookies that could not be decrypted anymore. I don't know for what sorts of data roundcube uses the encryption, but they never changed the default cipher method to not break backwards compatibility (see comment in `defaults.inc.php`) so I guess issues are to be expected from core roundcube as well. The instructions already contain a note regarding this. So I'd suggest to keep it that way to avoid breakage.

The other improvements (and stuff I missed in the migration steps that you pointed out) I will take over.

FingerlessGlov3s:
1. Yes I do have that tweak made actually. I forgot the HTTPS upgrade isn't default. Putting a note there seems like a good option.

2. Sounds good to me to do this option.

3. I did have an active session but I did browse to the root of roundcube. But that doesn't make a login happen. I don't think I test it in incognito... to see if the login screen loaded.

4. It was never set in the original mailcow docs the default will be the default one `DES-EDE3-CBC`. Changing it will only break current roundcube logged in sessions as the cipher is used to encrypt the IMAP password, nothing else. They do recommend changing it. This cipher is really weak, so we should definitely be changing it.  You don't stay logged in to round cub long anyway by default, so I don't see this as much of an issue. Just needs a note to say people will need to login again after the upgrade is done. Which if the upgrade takes a little while for them to do it, then they'll probably get logged out anyway.

mstilkerich:
I found some time to incorporate your feedback and update also the German version of the instructions. I hope it's good to go now!

MAGICCC:
Is this ready to merge then @mstilkerich and @FingerlessGlov3s?

FingerlessGlov3s:
I think so. 😀

mstilkerich:
Yes from my side it’s also ready to be merged. Thanks again @FingerlessGlov3s for testing everything and the great feedback!

MAGICCC (approved):

MAGICCC:
thx guys!

PR #580, opened 2023-06-02
Corrected roundcube password change guide

rallisf1:
Roundcube no longer parses plugin config files.
Adding the config parameters to the main configuration files fixes it.


MAGICCC:
Can you check this @mstilkerich and @FingerlessGlov3s? I do not use roundcube

mstilkerich:
For me roundcube parses plugin config files just fine, cannot reproduce this problem.

PR #581, opened 2023-06-07
Nextcloud Update via CLI added

deepc2020:
Added command for a cli based Nextcloud update to documentation. Both ways, interactive and non-interactive for scripts and cron, are described.


MAGICCC:
Thanks, but can you add an [admonition ](https://squidfunk.github.io/mkdocs-material/reference/admonitions/)before the part of automatic updates to warn users that it may break?

MAGICCC:
Will you implent the admonition @deepc2020?

MAGICCC:
Will you implent the admonition @deepc2020?

DerLinkman (approved):

DerLinkman (approved):
Thanks!

PR #583, opened 2023-06-13
Propose improve section on reverse proxy

raph-topo:
Preparing my contributions to reverse proxy topics (in particular Traefik v2 which I've now had 3 years production), I propose following new nav structure this section.


raph-topo:
Thank you for your comment!

PR #584, opened 2023-06-15
Create u_e-dovecot-sieve_filter.en.md

EricThi:


DerLinkman:
Don't know if we might add this as a new template in mailcow itself?

As we can create such templates.

MAGICCC:
Yes would makes sense

PR #586, opened 2023-06-21
Updating the Postgres DSN in Mailman3 Doc

asachs01:
After deploying the Mailman3 containers, I faced an error regarding the connection to the Postgres container. Digging into it, it appears that the sqlalchemy library doesn't support using `postgres` any more, but uses `postgresql` instead. This fixes the documentation.


MAGICCC (approved):

MAGICCC:
thx

PR #589, opened 2023-07-02
Update roundcube to 1.6.2

FingerlessGlov3s:
Update roundcube to 1.6.2, I'd also like someone else to test the upgrade too. I think I had an issue with the `composer update --no-dev -o` command, since my `composer.json` didn't exist but I think this is due to my setup.

If someone else like @mstilkerich could confirm there's no problems with their setup, I think we've fine to merge.

### DIFF
--- a/docs/third_party/roundcube/third_party-roundcube.de.md
+++ b/docs/third_party/roundcube/third_party-roundcube.de.md
@@ -45,7 +45,7 @@ Verzeichnis herunter und entpacken Sie es (hier `rc/`):
 
 ```bash
 mkdir -m 755 data/web/rc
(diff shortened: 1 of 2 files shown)


mstilkerich:
Hi,
actually I already upgraded roundcube to 1.6.2 following the instructions to verify they work. 

Upgrade worked without issues.

In fact the roundcube install script appears to run the composer update itself nowadays, so the extra step included in the instructions seems to not be needed anymore. However, it also does not hurt and I believe older versions of roundcube did not do the update. The release notes still mention it should be run when upgrading.

FingerlessGlov3s:
Sweet, so we're go for merge then. Shall we remove that step now or wait until 1.7 to double check it's going to carry on doing that?

mstilkerich:
As long as the roundcube instructions include the step I would keep it. It appears to be done since release 1.6.1.

https://github.com/roundcube/roundcubemail/issues/8642

PR #599, opened 2023-08-06
Some fixes and additions for Roundcube

sriccio:
- Added instructions to install and enable `dovecot_impersonate` plugin
- Fixed a thing in `mailbox.js` (btnSize is undefined)

### DIFF
--- a/docs/third_party/roundcube/third_party-roundcube.de.md
+++ b/docs/third_party/roundcube/third_party-roundcube.de.md
@@ -373,6 +373,25 @@ $MAILCOW_APPS = [
 
 Installieren Sie zunächst das Plugin [dovecot_impersonate](https://github.com/corbosman/dovecot_impersonate/) und fügen Sie Roundcube als App hinzu (siehe oben).
 
+```bash
+docker exec -it -w /web/rc/plugins $(docker ps -f name=php-fpm-mailcow -q) git clone https://github.com/corbosman/dovecot_impersonate.git
+```
+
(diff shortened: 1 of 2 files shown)


MAGICCC:
Can you check this @mstilkerich and @FingerlessGlov3s? I do not use roundcube, but looks fine-ish

mstilkerich:
I do not use this functionality, so I can’t comment based on experience. I would suggest to install via composer though instead of unpacking a git snapshot, like the other plugins in the documentation.

sriccio:
> Edit: Ok I see that this package is not registered on packagist and therefore installing via composer would require adding a custom repository to composer.json.
> 
> So to me this looks fine as it simply adds the shell commands on how to install and activate that plugin.

Hello,

Yes I recently went through the guide (very nice one btw) to add a RoundCube to an existing Mailcow installation.

Everything was perfect except it was missing command line to install the impersonate plugin. I searched around to find if it was installable via composer, but it is not.

So I went with git directly and was a bit skeptical first, as it seems the plugin had no update since years, but it worked wonderfully.

Without the impersonate plugin, when you impersonate the user from the mailcow admin interface, RoundCube is basing all the user profile on `user@domain*masteruser` instead of `user@domain` which create a lot of trouble.

So I thought it would be a good addition to have the commandline to install it easily.

Cheers,
Sébastien

MAGICCC:
Could you add the german part for it too, you can use deepl for the translation

sriccio:
> Could you add the german part for it too, you can use deepl for the translation

Done :)

MAGICCC (approved):

MAGICCC:
thx!

PR #700, opened 2024-04-18
Create Borg Backup even when backing up mysql

wb7:
Hello,

regarding this forum thread: https://community.mailcow.email/d/1796-borgmatic-does-not-backup-vmail/ Borgmatic isn't backing up everything when also dumping the database using a hook because of the `one_file_system: true` setting.

Adding the mounted volumes explicitly to the list solves this issue.

### DIFF
--- a/docs/third_party/borgmatic/third_party-borgmatic.de.md
+++ b/docs/third_party/borgmatic/third_party-borgmatic.de.md
@@ -73,7 +73,6 @@ Das nächste Kommando erzeugt dann die borgmatic-Konfigurationsdatei mit den kor
 ```bash
 cat <<EOF > data/conf/borgmatic/etc/config.yaml
 source_directories:
-    - /mnt/source
     - /mnt/source/vmail
     - /mnt/source/crypt
     - /mnt/source/redis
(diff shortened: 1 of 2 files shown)


mrclschstr:
Just for cross reference, I have already opened issues for this:

- https://github.com/borgmatic-collective/docker-borgmatic/issues/317
- https://github.com/modem7/docker-borgmatic/issues/125

MAGICCC:
So we can merge this safely?

mrclschstr:
Yes, I think that's a reasonable workaround for the moment. 

Maybe a small improvement: I'm pretty sure that the path `/mnt/source` in the borgmatic configuration under `source_directories` is no longer needed. Maybe @wb7 could confirm this briefly?

wb7:
I checked the suggestion of @mrclschstr and can confirm this; `/mnt/source` could be removed.

MAGICCC (approved):

MAGICCC:
ok thx

witten:
LGTM!

witten:
This should no longer be necessary, so it may be worth backing out or updating. Specifically, borgmatic 1.9.0+ no longer implicitly enables `one_file_system`  when databases are configured.

PR #723, opened 2024-06-10
Creating dhparam.pem as root (Troubleshooting > Common Problems)

Wave6677:
Mailcow's Dovecot container will not see dhparam.pem unless it is generated as root, and adding sudo will make it so any users running the command won't be confused as to why dovecot is still not running

### DIFF
--- a/docs/troubleshooting/debug-common_problems.de.md
+++ b/docs/troubleshooting/debug-common_problems.de.md
@@ -85,7 +85,7 @@ key.pem
 Wenn `dhparams.pem` fehlt, können Sie es mit Bash
 
 ```bash
-openssl dhparam -out data/assets/ssl/dhparams.pem 4096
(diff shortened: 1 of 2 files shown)


Wave6677:
> mailcow currently explicitly states, that you should do everything related to mailcow, directly with the root user...
> 
> [Link to docs](https://docs.mailcow.email/getstarted/install/#check-selinux-specifics:~:text=Please%20clone%20the%20repository%20as%20root%20user%20and%20also%20control%20the%20stack%20as%20root)
> 
> image

Yeah exactly, so the command itself should be run as root in some fashion, feel free to change it to remove the sudo and just state it needs to be done as root. A sysadmin is not going to remember a piece of installation documentation for one project they read over a year ago while theyre in the troubleshooting section trying to fix an issue.

Wave6677:
> At least updating should be done at least once a month.

update.sh tells you to run it as root.

PR #726, opened 2024-06-24
feat: add ACME rules to HAProxy config example

eastonman:
Updated the HAProxy config example. I think it is now more handy.

However, I need help with the German version because I don't know German.


DerLinkman (approved):
Merci! :)

PR #730, opened 2024-07-08
Add _submissions._tcp SRV record

SailReal:
See https://datatracker.ietf.org/doc/html/rfc8314#section-3.3 for further information


DerLinkman:
Merci :)

PR #733, opened 2024-07-15
Update minimum docker version requirements

Jniklas2:
Just updated the version numbers to the current major version. With the old `20.10.2`, mailcow doesn't start correctly anymore.

Maybe older versions (like version 25 or 26) are also working, I didn't test for the oldest working version.

### DIFF
--- a/docs/getstarted/install.de.md
+++ b/docs/getstarted/install.de.md
@@ -3,7 +3,7 @@
 
 ## Docker und Docker Compose Installation
 
-Sie benötigen Docker (eine Version >= `27.0.0` ist erforderlich) und Docker Compose (eine Version `>= 2.0` ist erforderlich).
+Sie benötigen Docker (eine Version >= `24.0.0` ist erforderlich) und Docker Compose (eine Version `>= 2.0` ist erforderlich).
 
 Erfahren Sie, wie Sie [Docker](https://docs.docker.com/install/) und [Docker Compose](https://docs.docker.com/compose/install/) installieren.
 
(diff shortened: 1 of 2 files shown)


DerLinkman:
27 is the newest Docker version?

I would not directly use this as the minimum version though.

It's even working with 24.X versions...

So i would say this is the minimum at least.

docs/getstarted/install.de.md
```diff
@@ -3,7 +3,7 @@
 
 ## Docker und Docker Compose Installation
 
-Sie benötigen Docker (eine Version >= `20.10.2` ist erforderlich) und Docker Compose (eine Version `>= 2.0` ist erforderlich).
+Sie benötigen Docker (eine Version >= `27.0.0` ist erforderlich) und Docker Compose (eine Version `>= 2.0` ist erforderlich).
```
DerLinkman, line 6: 27.X is the latest docker version i would not set this as a minumum as it is no minimum as it is the latest version existing.

For now lets set it to 24.X

docs/getstarted/install.en.md
```diff
@@ -2,7 +2,7 @@
     The installation is exactly the same on x86 and ARM64 platforms!
 
 ## Docker and Docker Compose Installation
-You need Docker (a version >= `20.10.2` is required) and Docker Compose (a version `>= 2.0` is required).
+You need Docker (a version >= `27.0.0` is required) and Docker Compose (a version `>= 2.0` is required).
```
DerLinkman, line 5: Here the same of course :)

DerLinkman (changes requested):
See comments

MAGICCC (approved):

PR #740, opened 2024-08-19
Authorize watchdog and bounces

dragoangel:
@DerLinkman can you please verify this PR and add DE translation?

Also if we already adding this to documentation we can change https://github.com/mailcow/mailcow-dockerized/blob/master/generate_config.sh#L417-L420 so it just include a link to the docs instead.


DerLinkman:
Yes! Will do it today

DerLinkman (approved):

PR #742, opened 2024-08-26
Remove legacy Nextcloud settings

MAGICCC:
As written [here](https://mailcow.email/posts/2024/release-2024-06/#discontinuation-of-nextcloud-script-and-support-for-installations-within-mailcow) mailcow will drop Nextcloud support end of 2024. This will be the PR to allow this


DerLinkman:
Hm... but inside the nextcloud docs page there is also a thing listed for the authentication against mailcow from nextcloud, that should be kept even after the removal of the nextcloud script.

MAGICCC:
Yes! Maybe we can create a site with different auths like we have for TOTP/fido

DerLinkman:
Merged it with https://github.com/mailcow/mailcow-dockerized-docs/commit/c1f689df604e9bc9a11097f2066499e0308ffb66

PR #762, opened 2024-10-02
Fix incorrect indentation

XSmeets:
This fixes a minor issue where a tab character was used instead of multiple spaces, leading to invalid YAML.


DerLinkman:
Is there a update needed to the german version too?

DerLinkman:
I see! Thank you!

PR #764, opened 2024-10-09
Improved "Forward the client network address to dovecot"

foorschtbar:
The original manual won't work, because the step of configure the plugin is missing AND the original plugin only support fixed addresses and no networks.

- Replaced roundcube-dovecot_client_ip with fork (with CIDR support)
- Added important configuration step to add trusted proxies (without this step it wont work!)


DerLinkman:
Looks fine to me!

DerLinkman (approved):

PR #769, opened 2024-10-15
Use python venv in actions

MAGICCC:


DerLinkman:
What's wrong?

MAGICCC:
Due to https://peps.python.org/pep-0668/ you cant use pip without venv anymore except you use apt install python3-pkg for every package we need.
And with `pipx` it's not possible to use something like `pipx install -r requirement`.

DerLinkman:
Ah yeah right, thought it would already run with venvs here...

PR #775, opened 2024-10-21
Update u_e-postfix-extra_cf.en.md

apio-sys:
Added an example of adding cipher restriction and TLS protocols to current standards.


DerLinkman (approved):
Merci! Corrected and checked!

PR #788, opened 2024-11-20
FIX: Update u_e-mailcow_ui-forgot_password.en.md

apio-sys:
Very small correction in the UI access instructions that made me search for a while for "Settings" where this should be "Options".


DerLinkman:
Merci!

PR #845, opened 2025-05-02
Add trusted proxies to nginx setup page

disgustipated:
Added note about TRUSTED_PROXIES that seems to have been added in the last release.
Apologies if is the German translation is poor, it is not my native language and I google translated it.


DasCanard:
+1

PR #848, opened 2025-05-10
Drop rsync.net in favor of another provider

purejava:
rsync.net won't work anymore with `docker-borgmatic` from May 16th

rsync.net informed me that:
> In addition, a v1.4 'borg' binary will be made available and can be reached by specifically calling '--remote-path=borg14'.  This means the list of 'borg' binaries that may be called is:
> 
> --remote-path=borg0     (borg v0.29) - available now
> --remote-path=borg12    (1.2.x branch - available now
> --remote-path=borg14    (1.4.x branch) - available after update
> 
> ... and we will be removing the plain old 'borg' binary which means you MUST explicitly specify a borg version using one of the three binaries above.  THIS MEANS that if you don't have a --remote-path argument in your borg command it will begin to fail on May 1 2025.

They have extended the deadline from 1st of May to May 16th.

Hetzner is a good alternative that can be pointed to in the docs. I already moved from rsync.net to Hetzner and the borgmatic backup is working well with Hetzner.

Further information can be found here:
https://docs.hetzner.com/de/storage/storage-box/backup-space-ssh-keys/
https://docs.hetzner.com/de/storage/storage-box/access/access-ssh-rsync-borg/

### DIFF
--- a/docs/third_party/borgmatic/third_party-borgmatic.de.md
+++ b/docs/third_party/borgmatic/third_party-borgmatic.de.md
@@ -74,7 +74,7 @@ source_directories:
     - /mnt/source/rspamd
     - /mnt/source/postfix
 repositories:
-    - path: ssh://user@rsync.net:22/./mailcow
+    - path: ssh://uXXXXX@uXXXXX.your-storagebox.de:23/./mailcow
       label: rsync
 exclude_patterns:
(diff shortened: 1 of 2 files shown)


DerLinkman (approved):
Thank you! Looks good.

PR #852, opened 2025-05-16
Add more SecuriteInfo database and remove spam_marketing from the list

dragoangel:
Actualize up-to-date list of useful SecuriteInfo, remove spam_marketing.ndb from the list and add warning about significant FPs from spam_marketing.ndb so it can be used only if user sure so.


accolon:
The following five rules you added are only available for _paid_ SecuriteInfo accounts:
securiteinfo0hour.hdb
securiteinfo.mdb
securiteinfo.yara
securiteinfo.pdb
securiteinfo.wdb

Since the mailcow docs currently focus on adding free databases / using a free account, this could lead to confusion.

dragoangel:
Hi @accolon thanks for catchup, then we can separate them a bit 😊

dragoangel:
@DerLinkman fixed

PR #859, opened 2025-05-27
add network_mode for ipv6nat disabling

jolly-jump:
yes. probably soon obsolete.
nontheless: if there is no network_mode, docker creates a bridge network "<foldername>_default" just for this part.
"host" is save, since it exists usually.


DerLinkman:
Obsoleted by Patch 2025-09

PR #872, opened 2025-08-07
Add Standalone Roundcube Installation (seperate Container)

CodeShellDev:
* added Integrated and Standalone for many Steps / Instructions using `=== "..."`
* updated existing Documentation to allow both Installation Types to be compatible with Plugins, etc.


CodeShellDev:
Fixes should now be available at PR #878

PR #879, opened 2025-08-15
Added Standalone Roundcube (german)

CodeShellDev:
Translated content from #872 into german (minus formatting errors)


CodeShellDev:
thank you!

PR #882, opened 2025-08-18
fix Roundcube (external)

CodeShellDev:
- cleaned up yml file
-  replaced wrong file in the german translation

Removed proxy network from docker-compose file which was left there from a previous draft.

Fixed german translation including the wrong site.roundcube.custom file


CodeShellDev:
included these fixes in new PR...

PR #885, opened 2025-08-22
Rework of Roundcube Documenation

CodeShellDev:
Hey there, this is another one of my _absolutely_ _infamous_ PRs ( :rofl: ).

### What did I do this time?

**_Well..._**
- to reduce confusion removed any mention of `ROUNDCUBEMAIL_PLUGINS` and `ROUNDCUBEMAIL_COMPOSER_PLUGINS` to instead prefer plugin installation via composer, since almost every other plugin is not installed traditionally
- removed `/web` / `data/web` from plugin documentation since user will know where config is located after having already finished the installation (user already edited the config file in the beginning where filepath is given) to reduce confusion between both installations
- **fixed roundcube smtp/imap host variables**

> _Decided that variables like smtp host, db host, db password should still be set in the environment for ease of access and other more advanced settings should be added in custom configs (but can also be set in the config only). Integrated Install's config is also mostly compatible with Standalone Install._

Basically this PR made me rethink the entire way that I was installing / managing Roundcube and due to that moving
the plugin installation into `inc.php` configs (like it should have been in the first place). 
Tho I decided against moving...

```yaml
      ROUNDCUBEMAIL_DB_TYPE: mysql
      ROUNDCUBEMAIL_DB_HOST: roundcube-db
      ROUNDCUBEMAIL_DB_USER: roundcube
      ROUNDCUBEMAIL_DB_PASSWORD: ${DBROUNDCUBE}
      ROUNDCUBEMAIL_DB_NAME: roundcubemail
      ROUNDCUBEMAIL_DEFAULT_HOST: dovecot
      ROUNDCUBEMAIL_SMTP_SERVER: postfix
      ROUNDCUBEMAIL_SMTP_PORT: 588
```
into the config.

> [!IMPORTANT]
> But that is for the maintainers to decide

> Personally I think setting these Variables inside of the Environment makes them more managable, but you could argue that moving this into the php config would be the last needed step to basically make the Integrated / Standalone Install identical (directory / file / structure wise). But this would kind of defeat the point of having the seperate Container...

### DIFF
--- a/docs/third_party/roundcube/third_party-roundcube.de.md
+++ b/docs/third_party/roundcube/third_party-roundcube.de.md
@@ -1,50 +1,39 @@
-## Installation von Roundcube
+# Roundcube installieren
 
-!!! note "Beachten Sie"
-    Sofern nicht abweichend angegeben wird für alle aufgeführten Kommandos angenommen, dass diese im mailcow
-    Installationsverzeichnis ausgeführt werden, d. h. dem Verzeichnis, welches `mailcow.conf` usw. enthält. Bitte führen Sie
-    die Kommandos nicht blind aus, sondern verstehen Sie was diese bewirken. Keines der Kommandos sollte einen Fehler
-    ausgeben; sollten Sie dennoch auf einen Fehler stoßen, beheben Sie diesen sofern notwendig bevor Sie mit den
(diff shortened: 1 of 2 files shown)


DerLinkman (approved):
Thank you!

PR #853, opened 2025-05-16
Add Roundcube Oauth Documentation (Base)

CodeShellDev:
- Seperated Installation into **Embedded** and **Standalone** for the above
- Removed confusing `cat <<EOCONFIG >`, which modifies php to add a `/` infront of every line (could confuse non-technicals or php newbies)
-  Added SSO and Oauth documentation for roundcube and mailcow

Added **Standalone** Installation to enable Roundcube and Oauth (and also to remove nesting)
You will now be able to follow the documentation to setup roundcube inside of a seperate docker container,
this will enable more complex environments to work and as a bonus it will walk you through, how to setup traefik as a reverse proxy.
Which I'd say is easier to setup and easier to manage, update and mantain.

Removed confusing `cat <<EOCONFIG >` and cleaned up code blocks.
`cat <<EOCONFIG >` seems to be not needed and for that was removed in my fork.

Lastly the Reason for this pull request:
Added documentation for setting up mailcow as oauth source for Roundcube...
Now you won't have to login twice,
1. in mailcow
2. in Roundcube

_Very sorry for the splitting part, not sure if it was a good decision, can also remove that..._

> [!IMPORTANT]
> This is the Source that #889 is based on
> Please continue in #889 instead

Related: [Community Post](https://community.mailcow.email/d/3675-sso-and-roundcube/13)

PR #889, opened 2025-09-14
Add Roundcube Oauth Documentation

CodeShellDev:
Added Documentation for using mailcow as Oauth Source for Roundcube, so that users can login via mailcow.

> [!IMPORTANT]
> This is a Continuation of #853.
> Since there were issues with the fork's master branch

### DIFF
--- a/docs/third_party/roundcube/third_party-roundcube.en.md
+++ b/docs/third_party/roundcube/third_party-roundcube.en.md
@@ -222,12 +222,12 @@ services:
 
 ## Standalone Install
 
-To Install Roundcube in its own Docker Container you have to add the following into your `docker-compose-override.yaml` file:
+To Install Roundcube in its own Docker Container you have to add the following into your `docker-compose.yaml` file:
 
 ```yaml
 services:
   roundcube-db:
-    image: mariadb:10.11
+    image: mariadb:10.11 # Set to the same version as mysql-mailcow to avoid having two mariadb images
     volumes:
       - roundcube-db:/var/lib/mysql/
     environment:
@@ -252,6 +252,8 @@ services:
       ROUNDCUBEMAIL_DB_USER: roundcube
       ROUNDCUBEMAIL_DB_PASSWORD: ${DBROUNDCUBE}
       ROUNDCUBEMAIL_DB_NAME: roundcubemail
+      ROUNDCUBEMAIL_SMTP_USER: '%u'
+      ROUNDCUBEMAIL_SMTP_PASS: '%p'
       ROUNDCUBEMAIL_DEFAULT_HOST: dovecot
       ROUNDCUBEMAIL_SMTP_SERVER: postfix
       ROUNDCUBEMAIL_SMTP_PORT: 588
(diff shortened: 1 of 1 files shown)


CodeShellDev:
**IMPORTANT UPDATE**

In the recent `v1.7` release of roundcube a new oauth method has been added `OAUTHBEARER` which seems to take overhand above the other methods (by default), so if your oauth flow is broken this is very likely the issue.

To fix this you have 2 options:
- **A** change the default back:

  ```php
  // Forced authentication type. One of "XOAUTH2", "OAUTHBEARER", "OAUTH" (default, auto-selection)
  $config['oauth_auth_type'] = "XOAUTH2";
  ```
 - or **B** leave the rc config as-is and instead go with `OAUTHBEARER`, for this you will need to update you dovecot config file:
   ```conf
   remote 172.22.1.0/24 {
     disable_plaintext_auth = no
   }
   remote fd4d:6169:6c63:6f77::/64 {
     disable_plaintext_auth = no
   }
  
   login_trusted_networks = 172.22.1.0/24 fd4d:6169:6c63:6f77::/64
   // add here
   auth_mechanisms = $auth_mechanisms oauthbearer
  
   passdb {
     driver = oauth2
     // and here
     mechanisms = oauthbearer
     args = /etc/dovecot/dovecot-oauth2.conf.ext
   }
  
   userdb {
     driver = static
     args = uid=vmail gid=vmail home=/var/vmail/%d/%n
   }
   ```

Hope this helps someone (did take a while to finally fix again)!

MAGICCC:
@FingerlessGlov3s possible to check that?

FingerlessGlov3s:
Not sure which way is best out of the two, I would think if we wanted to support `OAUTHBEARER` as a whole then those dovecot changes need to be done in the main dovecot config and I wonder if something is also needed for postfix.

Gmail looks to support only `XOAUTH2` at the moment, so that might be the better solution for now at least. I believe Thunderbird is `XOAUTH2` as well

CodeShellDev:
Right, this is meant for the roundcube x mailcow oauth flow, not as a whole.

CodeShellDev:
> Not sure which way is best out of the two, I would think if we wanted to support `OAUTHBEARER` as a whole then those dovecot changes need to be done in the main dovecot config and I wonder if something is also needed for postfix.
> 
> Gmail looks to support only `XOAUTH2` at the moment, so that might be the better solution for now at least. I believe Thunderbird is `XOAUTH2` as well

Only pro is that it seems to be the „newer“ method (and that it isn‘t made by google ;) )

FingerlessGlov3s:
I feel we're going to document a new oauth method for roundcube, we should see what the uptake of moving the postfix bit upstream so more applications can benefit from `OAUTHBEARER`, this way it keeps the roundcube documentation easier to follow and less prone to breaking changes from postfix updates.  Since `XOAUTH2` just works for the moment, it seems like the logical choose until mailcow stack officially provides OAUTHBEARER as well. Ideally the less changes needed is a better solution in my opinion. Makes it easier to support, and easier for users to follow.

CodeShellDev:
Makes sence, either way I have already given up on the idea of this being merged, was just so that someone stumbling onto this PR wouldn't make the same mistake!

PR #891, opened 2025-09-15
updated parameter to the used token instead of MAILCOW_EXPORTER_TOKEN

bastischubert:


MAGICCC:
Why can you specify the API key twice and why did you rename the variable?

MAGICCC (approved):

PR #892, opened 2025-09-20
proxy add mta-sts

crpb:
* add mta-sts example for apache, caddy, nginx and traefik


noahkw:
Thank you, the nginx part lgtm.

PR #898, opened 2025-10-06
Update third_party-roundcube.en.md

dudemcbacon:
The ROUNDCUBEMAIL_REQUEST_PATH env var is required otherwise roundcube will not forward POST requests to the correct path.


CodeShellDev:
> I just setup Roundcube for the first time and this env var was indeed required. Otherwise, the login was sending me to the active SOGo session.

Cool, I guess they may have fixed the bug, or it was just a problem with my setup (using traefik)!

CodeShellDev:
Had tried it out a few weeks ago, and it seemed to work with the environment variable!

MAGICCC:
@dudemcbacon Sorry can you add the line in the german version too?

PR #957, opened 2026-05-22
Update hardening for Postfix

apio-sys:
Since v1.11.0 from 2026-04-21 of Internet.nl, you will not be able to score 100% anymore with this procedure since this version includes the 2025-05 version of the NCSC TLS guidelines.  Without this updated, you will be sanctioned for insufficiently secure hash functions for key exchange (SHA1).

This updates the initial postfix hardening doc I posted in 2024 to reflect these new changes.

This is also related to issue [#7208](https://github.com/mailcow/mailcow-dockerized/issues/7208#issuecomment-4377264431) . The ciphers are hardened though as opposed to that proposed version to be better future ready. Also relates to external issue [#1553](https://github.com/internetstandards/Internet.nl/issues/1553#issuecomment-4313082980).

The German version has been machine translated so you might want to double-check this.

### DIFF
--- a/docs/manual-guides/Postfix/u_e-postfix-harden_ciphers.en.md
+++ b/docs/manual-guides/Postfix/u_e-postfix-harden_ciphers.en.md
@@ -16,3 +16,63 @@ smtpd_tls_mandatory_ciphers = high
 Such a configuration will pass current (2024-10-21) configuration checks against services like Internet.nl.
 
 If you want to adjust the ciphers for Dovecot as well you can find the corresponding tutorial [here](../Dovecot/u_e-dovecot-harden_ciphers.en.md).
+
+!!! warning "Important update Q2 2026"
+    Since v1.11.0 from 2026-04-21 of Internet.nl, you will not be able to score 100% anymore with this procedure since this version includes the 2025-05 version of the NCSC TLS guidelines. Please see below the extra configuration needed. Without this updated, you will be sanctioned for _insufficiently_ secure hash functions for key exchange (SHA1)
+
+Create a file `data/conf/postfix/openssl.cnf` with the following content:
+
+```
+postfix = postfix_settings
+
+[postfix_settings]
+ssl_conf = postfix_ssl_settings
+
+[postfix_ssl_settings]
+system_default = baseline_postfix_settings
+
+[baseline_postfix_settings]
+SignatureAlgorithms = ECDSA+SHA256:ECDSA+SHA384:ECDSA+SHA512:RSA-PSS+SHA256:RSA-PSS+SHA384:RSA-PSS+SHA512:RSA+SHA512:RSA+SHA256:RSA+SHA384
+```
+
+Add to your current `data/conf/postfix/extra.cf` the following:
+
+```
+tls_config_file = /opt/postfix/conf/openssl.cnf
+tls_config_name = postfix
+```
+
+So the complete `extra.cf` no may read:
+
+```
+tls_config_file = /opt/postfix/conf/openssl.cnf
+tls_config_name = postfix
+
+tls_high_cipherlist = ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256
+tls_preempt_cipherlist = yes
+
+smtp_tls_protocols = !SSLv2, !SSLv3, !TLSv1, !TLSv1.1
+smtp_tls_ciphers = high
+smtp_tls_mandatory_ciphers = high
+
+smtpd_tls_protocols = !SSLv2, !SSLv3, !TLSv1, !TLSv1.1
+smtpd_tls_ciphers = high
+smtpd_tls_mandatory_ciphers = high
+```
+
+for a properly hardened configuration. Beware not to override any existing settings you might had there if they are still needed.
+
+Restart `postfix-mailcow` to apply your changes:
+
+=== "docker compose (Plugin)"
+
+    ``` bash
+    docker compose restart postfix-mailcow
+    ```
+
+=== "docker-compose (Standalone)"
+
+    ``` bash
+    docker-compose restart postfix-mailcow
+    ```
+
--- a/docs/manual-guides/Postfix/u_e-postfix-harden_ciphers.en.md
+++ b/docs/manual-guides/Postfix/u_e-postfix-harden_ciphers.en.md
@@ -42,7 +42,7 @@ tls_config_file = /opt/postfix/conf/openssl.cnf
 tls_config_name = postfix
 ```
 
-So the complete `extra.cf` no may read:
+So the complete `extra.cf` now may read:
 
 ```
 tls_config_file = /opt/postfix/conf/openssl.cnf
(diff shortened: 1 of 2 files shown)


apio-sys:
Anychance of getting this merged si we van easily référence to this for hardening/PQC questions?

docs/manual-guides/Postfix/u_e-postfix-harden_ciphers.en.md
```diff
@@ -15,4 +15,64 @@ smtpd_tls_mandatory_ciphers = high
...
+So the complete `extra.cf` no may read:
+
+```
+tls_config_file = /opt/postfix/conf/openssl.cnf
+tls_config_name = postfix
+
+tls_high_cipherlist = ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256
+tls_preempt_cipherlist = yes
+
+smtp_tls_protocols = !SSLv2, !SSLv3, !TLSv1, !TLSv1.1
```
bobvandevijver: This line causes warnings during startup: 

```
/usr/sbin/postconf: warning: /opt/postfix/conf/main.cf, line 209: overriding earlier entry: smtp_tls_protocols=>=TLSv1.2
```

docs/manual-guides/Postfix/u_e-postfix-harden_ciphers.en.md
```diff
@@ -15,4 +15,64 @@ smtpd_tls_mandatory_ciphers = high
...
+smtp_tls_ciphers = high
+smtp_tls_mandatory_ciphers = high
+
+smtpd_tls_protocols = !SSLv2, !SSLv3, !TLSv1, !TLSv1.1
```
bobvandevijver: This line causes warnings during startup:

```
/usr/sbin/postconf: warning: /opt/postfix/conf/main.cf, line 213: overriding earlier entry: smtpd_tls_protocols=>=TLSv1.2
```

bobvandevijver (commented):
Two lines causes issues during startup. I've removed them from my configuration, as the default (`>=TLSv1.2) is equal to what is written in this documentation.

docs/manual-guides/Postfix/u_e-postfix-harden_ciphers.en.md
apio-sys: @bobvandevijver many thanks for testing this. The way the overrides in `extra.cf` works is described here; http://127.0.0.1:8000/manual-guides/Postfix/u_e-postfix-extra_cf/ . 

> Postfix will complain about duplicate values once after starting postfix-mailcow, this is intended.

So this can be ignored and are no issues as such.

docs/manual-guides/Postfix/u_e-postfix-harden_ciphers.en.md
apio-sys: See below, same reason.

docs/manual-guides/Postfix/u_e-postfix-harden_ciphers.en.md
bobvandevijver: Please explain how this added line, which causes warnings, improves the already present configuration which is `>=TLS1.2`. 

Warnings are intended behaviour to indicate that something is duplicated, yes. It shows you to investigate why that is, and if you've found a good reason, you can choose to ignore it. But here it is just not needed.

apio-sys: Sorry, my docs link from above is wrong and was pointing to my local dev. Here is the correct answer : https://docs.mailcow.email/manual-guides/Postfix/u_e-postfix-extra_cf/ . 

That said, you are right that these lines are now super-seeded by this commit https://github.com/mailcow/mailcow-dockerized/pull/5635 from over 2 years ago but got merged just recently Jun 26, 2026. My PR for this documentation update dates from May since this change was needed since well before if you wanted to harden properly. We can comment it out now with a note referring to the version where it is no longer needed.

Would be nice if a maintainer can feedback on what is wanted rather then leaving PR's lying about for months.

bobvandevijver: It got merged in 2024...

apio-sys: You are right, thanks. Indeed just after I committed the initial hardening documentation, it got included in the core thanks to @DerLinkman . So 2024-06   (2024-06-27) is the cutoff, and the override only ever mattered  for port 25 opportunistic TLS. I have removed those two lines from the updated documentation part. It should not overlap anymore. Paybe we want to get rid of the historic part I wrote in 2024. Maybe a maintainer can help decide what it to be kept.

apio-sys:
@MAGICCC any chance of merging this? The question comes up often enough  (at least in my environment). Thanks to the input of @bobvandevijver the doc is now as clean as possible without the double TLS version lines that were superseded in the 2024-06 release after my first version of this doc was published.

dragoangel:
Hi @apio-sys, we sure we want to add this as readme? I have doubts, when user will start to configure own openssl config - it will harder for mailcow maintainers to understand what changes safe to ship and what not. internet.nl mark a lot of stuff as _insufficiently secure_:

- SHA1 hash function
- DH-4096 Diffie-Hellman key exchange
- Cipher suite order - TLS_DHE_RSA_WITH_AES_256_GCM_SHA384 vs TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256
- Cipher suites

Some of this is definitely "oversecure" instructions that most of users will not get a win from. Would explain why:

SMTP is opportunistic by design, and having MTA-STS forces sender (who honors it) to not fallback beyond TLSv1.2 and force verification of certificate over root CA's, DANE just force TLS without version and verification done via TLSA record. Outgoing MTA who not support DANE/MTA-STS just try use StartTLS whenever this possible and when not - they fallback to plain text. Many mail server's over globe still can't do MTA-STS and not have DNSSEC enabled - as result their policy is "may", they also not updates quickly enough to get latest openssl, and may not support latest crypto to establish connection. While we can agree that most of internet already TLSv1.3 - you not want get 5% of hosts who would just fallback to plain text instead of some old cipher/DH over TLSv1.2 or fail with "can't establish ssl but required by dane" error.

My point: maybe better ship updated tls settings to default mailcow configs, but after careful review instead of instructing users heavily change their crypto related settings?

Cons: heavy changes requires alignment over time, many users may follow docs about change once - when they setup server or doing big maintenance, but not follow another (updated) docs over time, and just run update.sh - and this "secure" on first look change would become obsolete in 3 years, but users will stay on it.

apio-sys:
Hi @dragoangel . Thanks for your feedback. Much appreciated. Mailcow is shipped by default with somewhat "max compatible" settings which is good by default. This is just the documentation that helps those who want to go further in hardening. It's a subject regularly coming up and not very easy to deal with easily, hence this documentation update, from my initial version in 2024 that also helped mailcow to drop TLS <= 1.2 just a month after thanks to @DerLinkman . I think the updated version is valid to publish here since it clearly states what the odds are. 

> internet.nl mark a lot of stuff as insufficiently secure

and there are good reasons for that. More on their site for explanations (and yes, SHA1 should be dropped of course and since a very long time).

So either please help update the documentation if you have better ideas or help to get it merged so we can easily refer to it with those questions coming up on GH and the Forum. It's just one step in the next direction heading towards PQC.

@FreddleSpl0it @MAGICCC  or maybe other maintainers for the documentation repo, would appreciate some help in getting this finally merged (in a few months we can make the next version already). Or some feedback/discussion of course.

dragoangel (approved):

dragoangel:
Well, LGTM then, @MAGICCC @FreddleSpl0it btw I not have write access to docs repo :(

MAGICCC (approved):

apio-sys:
Thanks for helping merging this PR @MAGICCC but I think maybe went wrong since the changes are not visible here: https://docs.mailcow.email/manual-guides/Postfix/u_e-postfix-harden_ciphers/ . The deploy goes wrong here: https://github.com/mailcow/mailcow-dockerized-docs/actions/runs/32226255594/job/95986435850 .

MAGICCC:
thanks fixed.
Github issues as always

PR #967, opened 2026-08-01
Fix typo in Roundcube: 'docker-compose-override.yml'

CodeShellDev:
Small pr just noticed that technically the `*.override.yml` file does not work as-is, since the mailcow networks and volumes need to be attached too, so instead I propose we recommend adding it to the `docker-compose.yml` file.

### DIFF
--- a/docs/third_party/roundcube/third_party-roundcube.de.md
+++ b/docs/third_party/roundcube/third_party-roundcube.de.md
@@ -206,7 +206,7 @@ services:
 ```
 ## Standalone-Installation
 
-Um Roundcube in einem eigenen Docker-Container zu installieren, fügen Sie Folgendes in Ihre `docker-compose.yaml`-Datei ein:
+Um Roundcube in einem eigenen Docker-Container zu installieren, fügen Sie Folgendes in Ihre `docker-compose.override.yaml`-Datei ein:
 
 ```yaml
 services:
--- a/docs/third_party/roundcube/third_party-roundcube.de.md
+++ b/docs/third_party/roundcube/third_party-roundcube.de.md
@@ -206,7 +206,7 @@ services:
 ```
 ## Standalone-Installation
 
-Um Roundcube in einem eigenen Docker-Container zu installieren, fügen Sie Folgendes in Ihre `docker-compose-override.yaml`-Datei ein:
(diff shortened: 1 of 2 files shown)


Jniklas2:
I would recommend strongly against that, since the docker-compose.yml will get overwritten on every update. That's why the override files exist. It's even officially documented in the docker compose documentation: https://docs.docker.com/compose/how-tos/multiple-compose-files/merge/ (Keep in mind, you can use either docker-compose or compose as the file name and .yml or .yaml as extension).

CodeShellDev:
Ok, I guess that is a different opinion, will roll back but at least I am going to fix the typo…

Personally I think these override files are annoying and weird to deal with, since you have attach all of the needed volumes, networks etc. instead of just having them all in one place.

Jniklas2:
> Personally I think these override files are annoying and weird to deal with, since you have attach all of the needed volumes, networks etc. instead of just having them all in one place.

You only need to add things, you want to change or add. Since both simply get merged, you don't need to copy anything from the main file over to the override file. You can check the full config, that get's executed with `docker compose config`.

This is my override file, where I add an authentik ldap outpost to my stack and I don't need to define anything, that is already defined in the main file(like the network):

```yaml
services:
  authentik_ldap:
    image: ghcr.io/goauthentik/ldap:2026.5
    dns:
      - ${IPV4_NETWORK:-172.22.1}.254
    networks:
      mailcow-network:
        ipv4_address: ${IPV4_NETWORK:-172.22.1}.200
        aliases:
          - auth
    environment:
      - AUTHENTIK_HOST=[REDACTED]
      - AUTHENTIK_INSECURE=false
      - AUTHENTIK_TOKEN=[REDACTED]
```

MAGICCC:
yup indeed, thats why override exists. I'll close this for now

CodeShellDev:
Hey, there @MAGICCC, can we repurpose this PR to fix a typo in the file (docker-compose.override.yml)?
The name of it had somehow slipped under my radar: `docker-compose-override.yml` (`-` between `compose` and `override` should be a `.`).

MAGICCC (approved):

MAGICCC:
yep yep

PR #972, opened 2026-08-06
Update Traefik3 CertDumper path.

NKari69:
I'm correcting the destination path for certdumper following what I have on my server.

Remove the "expose" part for nginx as it's not need behind Traefik.

Add an alias for nginx inside the proxy network.

### DIFF
--- a/docs/post_installation/reverse-proxy/r_p-traefik3.en.md
+++ b/docs/post_installation/reverse-proxy/r_p-traefik3.en.md
@@ -84,7 +84,7 @@ SKIP_LETS_ENCRYPT=y
         volumes:
           - traefik_certs:/traefik:ro # mount your traefik certificate file
           - /var/run/docker.sock:/var/run/docker.sock:ro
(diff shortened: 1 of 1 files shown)


MAGICCC:
possible to check that @CodeShellDev, I dont use Traefik on my system, but the path is strange now
Also need the change on german version too @NKari69

CodeShellDev:
Hey there,
not sure what changed at certdumper‘s repository, but to me it seems like this update is not needed.
I am still running my setup without the `/domain.com` part of the mount. I guess @NKari69 wanted to separate the certs by domain…?
Depending on how mailcow loads the certificates this should be no different from the old one (IF it recursively reads the certs).

NKari69:
I simply have follow this guide. I think the path to the certificates had being created when I add the certdumper container.

CodeShellDev:
> I simply have follow this guide. I think the path to the certificates had being created when I add the certdumper container.

Sorry, could you clarify?

PR #974, opened 2026-08-16
Document how to fully disable outgoing TLS policy checks

dragoangel:
The TLS-Policy override guide only explained how to change the policy for a single destination. Where a firewall with SMTP inspection or an intercepting relay removes STARTTLS from the EHLO response, no per-domain entry helps and mail piles up in the queue.

Add a section to the English and German guide describing the extra.cf overrides that drop the postfix-tlspol socketmap and fall back to opportunistic TLS, including the scope of the change and the security implications.


dragoangel:
This document extension would cover questions asked in https://community.mailcow.email/d/3358-tls-is-required-but-was-not-offered-by-host-alt2gmail-smtp-inlgoogle

FreddleSpl0it (approved):

